Issue #26, opened 2014-10-07
Algorithm measurements


matthieu-vergne:
Considering that you are revising the conceptual modeling of your framework, I come here with an idea about how to model measurements that you could exploit in an experiment or simply as a way to get on the fly measures of algorithms (without being forced to use any experimental environment).

# Measures: Pull VS Push

First of all, measure makes no sense out of a specific algorithm. Speaking for instance about population size in an algorithm which does not manage a population of individuals makes no sense. Thus, I would delegate the definition of the measures to the specific algorithms. Consequently, when I speak about a measure I take the perspective of the algorithm being measured. From this perspective, I see two kinds of measures:
- *pull* measures, which come when requested by external entities (e.g. what is the current population size?)
- *push* measures, which come when generated by the algorithm (e.g. when are new individuals generated?)

Basically, the algorithm have no idea of when a *pull* can come, as it is decided depending on the needs of the external entities. At the opposite, the algorithm is the master when it comes to a *push*, because it is the one which knows when such a measure is feasible and generate it consequently.

Thinking about the code, I would model that with 3 interfaces:
```java
public interface Measure {
	// Name of the measure, e.g. "population size"
	public String getName();
	// Description of the measure, e.g. "Provide the current size of the population of individuals."
	public String getDescription();
}
```
This generic interface aims at centralizing the common "administrative" methods that a measure should provide. If you want a measure to be freed from these administrative work, then you can let the generic interface empty (but still present) and put these methods in an external interface like `NamedEntity`. This way, not only you can let the algorithms choose whether they implement or not this stuff (it is mainly needed for GUIs, so it is arguable to force a measure to implement them) and reuse it for other stuff (like algorithms themselves, problems, and so on, which would make sense for GUIs too). Thus, my recommendation would be the separation.

Then, extending this generic interface:
```java
public interface PullMeasure<T> extends Measure {
	// Request immediately the current value of the measure
	public T get();
}
```
```java
public interface PushMeasure<T> extends Measure {
	// Register to obtain the measure when it is computed
	public void register(MeasureListener<T> listener);
	// Unregister to free the measure when it is not needed anymore
	public void unregister(MeasureListener<T> listener);
}
```
The former is trivial, but the latter uses a typical event-based pattern: the listener. You find it for instance extensively in Swing to manage the GUI's events (mouse click, keyboard button pushed, etc.). Thus, we come with this additional interface:
```java
public interface MeasureListener<T> {
	public void measureGenerated(T value);
}
```

[...]

PR #45, opened 2015-01-30
First Generic Measure Implementations

matthieu-vergne:
Here you have the `SimplePushMeasure` and `SimplePullMeasure`, which provide all the basic implementations for a basic push/pull measure, and the `LastEvaluationMeasure` and `DurationMeasure` which extend them to have the equivalents of what I presented in #26 .

For the population measure, just extends `SimplePullMeasure` and implement the `get()` method.

### DIFF
--- a/jmetal-core/pom.xml
+++ b/jmetal-core/pom.xml
@@ -68,5 +68,10 @@
             <artifactId>hamcrest-all</artifactId>
             <version>1.3</version>
         </dependency>
+		<dependency>
+			<groupId>fr.matthieu-vergne</groupId>
+			<artifactId>data-access</artifactId>
+			<version>1.1</version>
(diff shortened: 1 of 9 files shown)


matthieu-vergne:
Notice that I did not introduced tests for `SimplePullMeasure` because it is abstract. We could make it non-abstract by adding an interface (like the listener is an additional interface to the `PushMeasure`) which provides the value to return. You can look at the implementation of my `Puller` and its `ValueGenerator` to get my point. This is exactly what it does.

matthieu-vergne:
Notice that none of the implementation introduced through this pull request use my lib. As said in #26 they are just too simple to justify the integration effort, at least if you want to have your own interfaces to ensure independency.

PR #46, opened 2015-01-31
Factoring + Counting Measure + MeasureManager

matthieu-vergne:
I factored the two generic implementations and I have added the generic `CountingMeasure`, which allows:
- to count occurrences manually by calling `increment()`
- to count occurrences automatically from any `PushMeasure` by registering to it through `link(measure)`

This way, you can use it to count how many evaluations have been made by linking it to an instance of `LastEvaluationMeasure`:
```java
LastEvaluationMeasure<Solution, Value> lastEvaluation = new LastEvaluationMeasure<>();

CountingMeasure evaluationCounting = new CountingMeasure();
evaluationCounting.setName("Evaluation counter");
evaluationCounting.setDescription("Count the number of evaluation made by the algorithm.");
evaluationCounting.link(lastEvaluation);

// algorithm execution

long evaluationDone = evaluationCounting.get();
```
This is a basic example of building a measure based on another. More generally, during an experiment with a GUI for instance, you can identify the instances of `PushMeasure` provided by an algorithm, list them to the user, allow him to select which measures should be counted by checking some checkboxes, and create a `CountingMeasure` for each of them that way (create+link). And because the `CountingMeasure` is also a `PushMeasure`, you can register to it to update the screen in real time during the experiment. It is also a `PullMeasure`, so you can get the value on demand too, for instance if you want to take a snapshot at a given time of the execution.


matthieu-vergne:
I have completed the Javadoc for all the measurement interfaces/classes + added an implementation for the `MeasureManager`.

ajnebro:
About the `LastEvaluationMeasure` class, it does more than I had expected :). In fact, I see that class `CountingMeasure` just could fit into the my former idea of just knowing the current number of evaluations. 

I have added a the method:
```java
  /**
   * Increment the current count in a given amount
   */
  public void increment(int amount) {
    count += amount;
    push(count);
  }
```
thinking about how incorporate the code in the algorithms. For example, in NSGAII the natural places seems to be these methods:
```java
  @Override protected void initProgress() {
    iterations = 1;
  }

  @Override protected void updateProgress() {
    iterations++;
  }
```
Here I'm using iterations instead of evaluations, but I could have chosen:

  @Override protected void initProgress() {
    evaluations = population.size();
  }

  @Override protected void updateProgress() {
    evaluations += population.size() ;
  }

ajnebro:
An advantage of using pull and push measures is avoiding having to modify the code of the algorithms whenever we need to use the measures. When we are running the algorithms for performance assessment we need no measures at all, but if we use them to solve a complex problem and the computation can take hours or days, then measures can be very useful. Adding and removing code in many algorithms to cope with these two usages is bored, cumbersome and error prone.

The point now is to decide how incorporate the measures in the algorithms in a let's say structured and elegante way. We could use NSGAII and the NSGAIIRunner class to test ideas.

matthieu-vergne:
When you speak about the combination of `LastEvaluationMeasure` and `CountingMeasure` to build a measure which counts the number of evaluations, I would say that it was my first objective, so nothing surprising. Maybe I forgot to say it explicitely, though.

But then I don't follow you, because in the example of code you show with `init/updateProgress()`, not only it is about *iterations* instead of *evaluations*, but also the fact that you use a simple variable instead of the `CountingMeasure`, so I don't really get the point.

Regarding the mix variable/measure: if you intend to use the measure, then instantiate the measure in `initProgress()` (or better, instantiate it in the constructor and reset it in `initProgress()`) and use `measure.increment()` in `updateProgress()`. If you want to use the variable, then use as a measure `SimplePullMeasure` and, for the implementation of its `get()` method, just read the variable. You reuse your variable, but you lose the reactivity of the `PushMeasure`.

Regarding the mix iteration/evaluation: an evaluation can occur at anytime, and even if you make one or a constant amount of evaluations per round, it is rarely a reliable measure for round counting. Because semantically, you count a round when an iteration starts or finishes, so you would use `CountingMeasure.increment()` as the very first or very last line of a loop (or directly in its definition, like:
```java
for(CountingMeasure r = new CountingMeasure ; r.get < roundLimit ; r.increment()) {
    // ...
    // nothing here about r unless the algorithm needs r.get()
    // ...
}
```
The evaluation, you would use it *during* the iteration. Even if it is close to the end of the loop, the point of being notified about the end of the iteration is to know that **all** what should have been computed during the iteration have been computed, so nothing more should be done during this iteration. It could be that you make your evaluation at some points in the end, and then decide about some additional stuff to do with your solution, so that when the evaluation is done you still have stuff to do, like adding the solution to the population or not. If an external user relies on the evaluation notification to say "OK, so let's get the population resulting from this iteration", then he misses the point, because after your notification you will modify this population, so the one he got is not completely representative of the finished iteration. I don't know if this mix you did was because of such misalignment or just a coincidence, but in any case I would recommend you to not link evaluation counting and iteration counting. If the point is that you could use the `LastEvaluationMeasure` in a tricky way to count iterations, then don't forget that this measure is nothing more than a decorated extension of the `SimplePushMeasure`, so prefer to use the `SimplePushMeasure` for notifying the end of an iteration, although the `CountingMeasure` used manually with `increment()` does that and more.

matthieu-vergne:
Regarding the integration of the measures:
> An advantage of using pull and push measures is avoiding having to modify the code of the algorithms whenever we need to use the measures. When we are running the algorithms for performance assessment we need no measures at all, but if we use them to solve a complex problem and the computation can take hours or days, then measures can be very useful. Adding and removing code in many algorithms to cope with these two usages is bored, cumbersome and error prone.
> 
> The point now is to decide how incorporate the measures in the algorithms in a let's say structured and elegante way. We could use NSGAII and the NSGAIIRunner class to test ideas.

If I take it literaly, you are wrong:
> When we are running the algorithms for performance assessment we need no measures at all, but if we use them to solve a complex problem and the computation can take hours or days, then measures can be very useful

Within an algorithm, you already use "measures" (in the broad sense) implicitly, because you store values in variables for future reuse during the running of the algorithm, to take decisions and so on. If no measure, no decision, and if no decision, no algorithm at all. The only point is to know which of them you want to make available from outside the algorithm (by providing them through a `MeasureManager`), so external users can read/be notified about them.

The duty of the algorithm is to create the `Measure` instances it needs internally and make them available for external entities. Then, because you make these measures available externally, other entities can not only get them but also build on them to add other interesting measures for their own context. Thus, an algorithm could provide notifications through `LastEvaluationMeasure` only and an external entity could link it to a `CountingMeasure` to have an additional measure. Then the point is to see which strategy to use in order to build on available measures, and this relates to my suggestion of listing the available measures to the user and letting him decide for instance which one he wants to add a counter (for the specific case of `CountingMeasure`) or anything else (if we implement other measures able to "link" to existing ones). Now, this kind of strategy has to be made depending on the context, like an experiment. So if I have well understood, the `NSGAIIRunner`, although I would see something more generic than an algorithm-dedicated runner.

ajnebro:
> But then I don't follow you, because in the example of code you show with `init/updateProgress()`, not only it is about *iterations* instead of *evaluations*, but also the fact that you use a simple variable instead of the `CountingMeasure`, so I don't really get the point.

Yes, I included the current code to indicate where to used the measure. It should be properly modified.

matthieu-vergne:
OK. Now I understand better. {^_^}

ajnebro:
> If I take it literaly, you are wrong:
>> When we are running the algorithms for performance assessment we need no measures at all, but if we use them to solve a complex problem and the computation can take hours or days, then measures can be very useful
>
> Within an algorithm, you already use "measures" (in the broad sense) implicitly, because you store values in variables for future reuse during the running of the algorithm, to take decisions and so on. If no measure, no decision, and if no decision, no algorithm at all. The only point is to know which of them you want to make available from outside the algorithm (by providing them through a `MeasureManager`), so external users can read/be notified about them.

Hmmmmm, yes, I think I get the point: the algorithm designer has to decide which measures to provide to the user, who will have the choice of using none, some or all of them.

ajnebro:
Then, if measures are a part of the algorithm, performance costs should be taken into accout, mainly in  push measures. 

For example, computing the current evaluation can be negligible, but let's think on a feature that frequently have been raised to me by some people: how to write into files the population in each iteration. If we use a pull measure to get the current population (to then write it into files), the performance will be conditioned by the frequency of the pulls, but this is up to the user. 

If the user uses a push measure instead, perhaps receiving the population at the end of each iteration could be excesive. Then we should consider to add a kind of parameter to decide the frequency of the pushs, or a condition to raise them.

Taking again NSGAII as an example, the main measures would be:
- current iteration or evaluation (Integer) 
- current running time (Long)
- current population (List<Solution>)

but others could be considered:
- number of non-dominated solutions in the population (Integer)
- non-dominated solutions in the population (List<Solution>)
- number of solutions not violating side constraints (Integer)
- solutions in th population not violating side constraints (List<Solution>)

matthieu-vergne:
> how to write into files the population in each iteration

Well, if ths objective is exactly that, now you know how to do it: the algorithm provides a push measure to notify when an iteration finishes (`CountingMeasure` being a good candidate) and a pull measure which provides the current population on demand. Otherwise you create measures which build on the avaialble ones to obtain this information. Then, users register to the push measure to know when the end of the iteration occur. In this listener, they call the pull measure's `get()` to know what is the current population and they write it into a file.

If you want to make it more performant by exploiting multi-threading, then you store the current population into a separate collection and run a new thread which will feed the file. Going even further, you can use this listener only to store the collection into a list (to preserve the order) and wait the end of the execution to write all the collections stored. Meanwhile, you can use the stored collections to display them on the screen for the user to check them on real time. You can imagine what you want.

Regarding the performance cost, there is a part which is relevant to the algorithm itself, and another part which is relevant to the external users (and the algorithm should not care about). Which structure to use in order to store your population during the running is an algorithm-specific performance issue, and it should be chosen depending on the accesses you do to this structure. How long an external listener takes to manage the pushed value is an external issue. If the user register a heavy computing listener, the algorithm should consider that it is because he wants it. If he does not, then he has to refine it to obtain what he wants. This is not the job of the algorithm.

matthieu-vergne:
> If the user uses a push measure instead, perhaps receiving the population at the end of each iteration could be excesive. Then we should consider to add a kind of parameter to decide the frequency of the pushs, or a condition to raise them.

Then you objective is not to get the population at each iteration anymore. But anyway, it is up to the user who register: if you don't need each iteration but only every X iterations, then still register to the listener, but each time the listener is called it increments a counter, and when the limit is reached it get the population and write it before to reset the counter, instead of doing it at each notification. If you want to use a frequency, then you should run a thread which will wake up every X seconds and call the get() method of the pull measure, no need to rely on the push measure because the frequency should not be controlled by the algorithm.

matthieu-vergne:
Regarding the running time, I think that there is the algorithm and the user perspective:
- the algorithm has a time limit, so it should not take more time than the limit.
- the user want to calculate the time consumed by the algorithm

For the algorithm, it will use a time duration which entails eerything, from the start to the end, because it wants to compute not more than the limit, whatever the computation contains. For the user, the time start at the launch of the algorithm and stops at the end, but should be also paused when some parts specifically added by the user are called, because these parts are not part of the algorithm and thus should not be counted as "algorithm-dedicated" time.

Each of them can use their own measure of duration for that. And in this case, each of them have full control of the specific time they want to measure. One is an algorithm-specific measure which compute an algorithm-specific duration, the other is a user-specific measure added by him to compute a user-specific duration. Once again, it is a matter of responsibility, and this is from this responsibilities that you identify who should instantiate and control the measure.

ajnebro:
The algorithm perspective has sense when the time limit is in the stopping condition (instead of performing a fixed number of iterations/evaluations) but, as I see it, it is not a measure.

The user may want to know the time consumed by the algorithm by a number of reasons: merely as a curiosity, to know (after getting the current population and some how analyzing it) about the search progress at this moment, etc. This perspective is clearly a measure to me.

matthieu-vergne:
> The algorithm perspective has sense when the time limit is in the stopping condition (instead of performing a fixed number of iterations/evaluations)

Yes, and more generally when the algorithm itself use the computation time to take decisions, whatever it is for (stopping, allocating resources, etc.).

> but, as I see it, it is not a measure.

In the broad sense it is, as discussed before, now it could be that you don't care about making it a `Measure` instance, because you don't plan to make it available outside. That is OK. Now, you could compute it your own way or exploit a `Measure` implementation which already provides what you need to deal with that. This is an implementation decision which has to be made in the scope of the algorithm, independently of how the algorithm will be used. It is not because you don't plan to make it available outside that you cannot use any `Measure` feature, because it is anyway a measure, although an internal-only one.

> The user may want to know the time consumed by the algorithm by a number of reasons: merely as a curiosity, to know (after getting the current population and some how analyzing it) about the search progress at this moment, etc. This perspective is clearly a measure to me.

In that case, and this is what I wanted to say in my previous post, I think that this is a measure which has to be managed *by the user* directly, not the algorithm. The user has to instantiate it, start it when he starts the algorithm, stop it when the algorithm stops, but also pause it when some additional parts (set by the user to help him evaluate the algorithm) consume the computation time. For instance, the user could want to know the execution time of a single run of the algorithm, or an average on X runs, or the average time between two evaluations, etc. And you will not compute the same value depending on that, yet these are all time measures. The important point here is that these are **user** measures, not **algorithm** measures, so there is no reason to count on the algorithm to provide any of them. This is the user which should create them and they will disappear when the user has finished, even if the algorithm is reused after for another purpose.

# Example of "runner": average duration to reach a given value

I am currently implementing a "runner" (I don't know if it fits your conception of runner, but at least it is a class aiming to run an algorithm). It is a simple time performance analyser. The goal here is to make several runs of an algorithm in order to know how long it takes to reach a given level of quality. It is initialized that way:
```java
algorithm = /* algorithm to run, provided in argument */;
qualityGoal = /* value to reach, provided in argument */;
qualityMeasure = /* measure to get the current value, provided in argument */;

// I have introduced a Configuration type. You can consider it
// like a structure which allows you to apply a specific set of
// values to a specific set of parameters.
runConfiguration = /* configuration to run the algorithm, provided in argument */;
stopConfiguration = /* configuration to force the algorithm to stop, provided in argument */;

// instantiate the duration measure we will use to compute
// the specific time that we want to measure
durationMeasure = new DurationMeasure();

qualityMeasure.register(new MeasureListener<QualityValue>() {

	@Override
	public void measureGenerated(QualityValue value) {
		// we stop the time counting because we start a
		// user specific computation
		durationMeasure.stop();
		
		if (value.compareTo(qualityGoal) >= 0) {
			// quality goal reached: we can stop the algorithm,
			// so we apply a configuration aimed to stop the
			// algorithm as soon as possible, it could be
			// immediate if the algorithm support it or it could
			// be a configuration which trick the algorithm to
			// stop the next time it checks its stopping
			// condition.
			stopConfiguration.applyAll();
		} else {
			// quality goal not reached yet
		}
		
		// we restart the time counting because the
		// user specific computation is over
		durationMeasure.start();
	}
});
```
So here you see that the measure is introduced by the analyser itself, such that it is configured/used in a manner to measure exactly what it wants to measure. Another part of the initialization is a parallel thread which displays (during the run of the algorithm) the current level of quality achieved so far:
```java
Thread regularDisplayThread = new Thread(new Runnable() {

	@Override
	public void run() {
		// start a loop which will stop when the complete analysis will be
		// finished.
		while (isRunning[analysis]) {
			try {
				// wait a "run start" notification
				wait();
				
				// during the run, display every seconds the
				// quality achieved so far.
				while (isRunning[round]) {
					// display the current quality
					System.out.println("So far: " + qualityMeasure.get());
					
					// wait 1s or a "run stopped" notification
					wait(1000);
				}
			} catch (InterruptedException e) {
				e.printStackTrace();
			}
		}
	}
});
```
While this thread run in parallel, the main loop is like that:
```java
// collection to store the durations of the different
// runs for average computing.
Collection<Long> durations = new LinkedList<>();

// Make 100 runs to compute the average
for (int i = 0; i < 100; i++) {
	// configure the algorithm to be ready to run, especially
	// because at the end of each run, it is configured to be
	// stopped as soon as possible.
	runningConfiguration.applyAll();

	// tells "run started" to the display thread
	isRunning[round] = true;
	regularDisplayThread.notify();

	// run the algorithm and measure its duration
	durationMeasure.reset();
	durationMeasure.start();
	algorithm.run();
	durationMeasure.stop();

	// tells "run stopped" to the display thread
	isRunning[round] = false;
	regularDisplayThread.notify();

	// retrieve the run duration
	Long currentDuration = durationMeasure.get();
	durations.add(currentDuration);

	// compute the average duration with previous runs
	double average = 0;
	for (long duration : durations) {
		average += duration;
	}
	average /= durations.size();
	
	// display the results before to restart a run
	System.out.println("Time spent in run " + i + ": "
			+ currentDuration + " (AVG: " + average + ")");
}
```
I never run it, I am still implementing it, but the only assumptions so far are the following:
- the algorithm implements `Runnable` (to call `algorithm.run()`), I don't need it to implement `Parameterable`, `Measurable` or even `Algorithm`, because all the measures and parameters (through configurations) are provided in arguments, so they can be built in whatever manner (provided by the algorithm or implemented from scratch to adapt an algorithm implemented out of the jMetal framework, like creating a `Parameter` for each setter and a `PullMeasure` for each getter)
- the quality measure is at least a `PushMeasure` (for applying the stopping configuration), the `PullMeasure` is added if it is not present (for the regular display thread)
- the quality goal has a type compatible with the quality measure (so I can compare them)
- the type of the quality values extends `Comparable` (so I can use `compareTo()` to make this comparison), which is typically the case for `Integer`, `Double`, `BigDecimal` and many others

The two points I would like to highlight here are that:
- this analyser is independent of the algorithm evaluted (it relies on generic assumptions)
- the duration measure it uses is entirely managed by the analyser, without relying on any duration measure provided by the algorithm

PR #50, opened 2015-02-01
Reset Counting + Fixes + New Measure Implementations

matthieu-vergne:


matthieu-vergne:
I have added a factory (`MeasureFactory`) which has one method to create a `PullMeasure` based on a `PushMeasure`, it store the pushed values in order to retrieve them at any time with the pull, and another to create a `PushMeasure` from a `PullMeasure`, it checks at regular intervals (customizable) the pull and notifies with the push in case of modification.

For the last one, there is advanced coding to manage the regular check, so you could be interested to look at it for optimization purpose:
- the use of a `Thread` to run the check in parallel
- the use of `Thread.sleep()` to free the processor until the next check
- the use of a `WeakReference` to ensure that, when the `PullMeasure` can be destroyed by the garbage collector, we don't block it by keeping a reference on it, which would lead to have a thread running forever and memory leaks.

Probably it could be further improved by using a single thread for all the measures created that way, rather than running a single thread for each of them. I added a todo in the method for that, with suggestions.

You can also look at the tests, where I put some comments to explain what/how we test it:
- `MeasureFactoryTest.testCreatePushFromPullNotifiesWithTheCorrectFrequency()`
- `MeasureFactoryTest.testCreatePushFromPullNotifiesOnlyWhenValueChanged()`
- `MeasureFactoryTest.testCreatePushFromPullStopNotificationsWhenPullDestroyed()`

Additionally, there is also a `PullPushMeasure` which combines both the abilities. It is adapted for the creation of new measures as well as wrapping existing measures.

With all that, we can switch easily from pull to push and vice-versa, and even combine them.

matthieu-vergne:
Now we also have what we need to build measures automatically from objects coming from anywhere, based on the (public) getters and members of the classes. This should be useful to apply the experiment stuff that we will implement to any algorithm, even built without using the jMetal framework. By displaying the measures built automatically (e.g. on a GUI), the user can decide which one to use without having to code anything.

PR #51, opened 2015-02-02
Fix: background checks still running when push measure destroyed.

matthieu-vergne:
For the periodical check made on the pull measure to feed the push measure, the thread was stopped when the pull measure was destroyed (no change can happen anymore, so no need to recheck), but not when the push one was destroyed (nothing to feed anymore, so no need to recheck). Now both cases are considered.


ajnebro:
Matthieu,  
  when trying to compile the project I get errors relating to these classes:
```java
import org.uma.jmetal.util.generic.DescribedEntity;
import org.uma.jmetal.util.generic.impl.SimpleDescribedEntity;
```
Please, could you check whether they are added to git?

matthieu-vergne:
Sorry, should be fixed in #52 : I made them when introducing the parameters, but I did not push parameters stuff (they are on my experimental branch). I rebased my history so these imports came here while the corresponding stuff did not. The pull request add the classes so it should compile.

PR #52, opened 2015-02-02
Add DescribedEntity to factor name and description.

matthieu-vergne:


ajnebro:
Thanks :). Now everything compiles fine.

matthieu-vergne:
Don't hesitate to use `DescribedEntity` and `SimpleDescribedEntity` for factoring existing stuff, like algorithms name/description and stuff like that (if there is).

ajnebro:
Sure, I will make us of them. The only point is that I would use as package name something like "naming" or "description" instead of "generic".

matthieu-vergne:
Sure, just move the stuff. Even for the measures, considering that it is central for the genericity of the algorithms, it should not be in "util". There is some reorganization to do their.

PR #53, opened 2015-02-10
list junit and mockito as test dependency

deric:
jUnit and mockito depedencies should be in scope `test`, otherwise are required as runtime dependencies, which is really not necessary.


ajnebro:
Thanks :)

PR #56, opened 2015-03-17
Fix the generic types to delegate the casting work to the manager.

matthieu-vergne:
This is to simplify the use of the manager. Otherwise, you always need to cast the result to make something with it. I have modified the runner consequently.


matthieu-vergne:
Notice also that there is already a generic class for having both pull and push features. I have noticed an issue regarding its use (I have added a FIXME for that) but your SingleValueMeasure is a duplicate. Probably it would be better to specialize mine for the enclosing of existing measures and keep yours for creating a brand new one. Think about moving your class in the `measure.impl` package, considering how generic it is.

PR #62, opened 2015-05-12
Jmetal45

GiovaniGuizzo:
I made 3 main modifications in this version.
### Normalization

I identified a problem with the indicators regarding how the normalization was made.

Before my modifications, the maximum and minimum values were obtained using the true Pareto front for some indicators (such as hypervolume). However, opposed to the minimum value, the maximum value of an objective function does not necessarily is present in a solution of the true Pareto front. For instance, to calculate hypervolume, take the true Pareto front: `{{2, 7}, {3, 5}}` and a dominated solution `A {12, 9}`. When the solution A would be normalized, the previous normalization would do: `(12 - 2) / (3 - 2) = 10` for objective 0 and `(9 - 5) / (7 - 5) = 2` for objective 1. Then a "normalized" solution A with the values `{10, 2}` would be used to calculate the hypervolume of a Pareto front.

To overcome this limitation, I added to 3 indicators (HV, IGD and GD) a new method in which the user can inform as parameter the arrays of maximum and minimum values. Therefore, the user can give the real maximum and minimum (12 and 9 for the example) and then the normalization can be performed correctly. I preserved the old methods.

I highly recommend to take a look on how the normalization is done everywhere because I don't know where the normalization is done this way. I also am concerned about this issue on the new version of jMetal.
### Calculator Classes

I added three calculator classes to ease the computing of hypervolume, IGD and GD because of the normalization problem mentioned before. Briefly, these calculator classes must store the Pareto fronts which will be compared in an internal population. At the end, after storing the fronts, the user invokes the execute method for each Pareto front and these calculators use the internal population (which has all solutions of the comparison) to find the maximum and minimum values.

The classes are annotated for more info. It is not a "must" change as the normalization issue, but merely an addition of functionality which may come in handy. It may not be the best way to do this calculation or to find the maximum and minimum values for a bunch of fronts, but it is easy to understand (at least for me) and works well.
### Added Methods

I added two methods which I use very often: a method in MetricsUtil class to get the non-dominated solutions of a solution set, and a method in NonDominatedSolutionList to add all the solutions of a solution set.


Juanjdurillo:
Hi Giovani, 

interesting discussion. In my view (and it does not have to be the right one) I am kind of biased for the normalization we have been doing so far (although I am open to modifications if you show me is wrong). I focus on your example: You will be calculating the HV of the point {10, 2}, and as this point will be dominated by the reference point used by the HV (1,1), its HV will be 0 (which in my view is not wrong, since the point is really bad. 
Focus now on the following example: the true Pareto front is {2, 1}, and {1, 2} and the approximation front is {{100000000, 1},{55,55},{1, 100000000}}. Obviously the Pareto front dominates the other front (which is indeed a really bad approximation to the Pareto front), using the current approach, the dominated front will report a value of HV = 0.0. Using the approach you suggest (or I have understood you suggest), will require normalizing the front using the maximum values 1000000000, 10000000. If you compute the HV of such a front using this approach, the HV reports the value 0.9999989200002808, which seems to be close to 1 and give the idea of a good approximation. However, this is contradictory since we know that it is a really bad front.  

We normalize using only the Pareto front because we want to scale an approximation to the Pareto front. We do not want to have everything normalized, this is not the goal. The goal is to scale one front to the other so we are comparing the same thing. If, after normalization we get points which are not in 0,1 for the approximation, that is fine (the point was actually not very good), and will be reflected by the quality indicators. But again, this is my view.

GiovaniGuizzo:
Hello Juan.

Indeed, it is a interesting discussion.

I understand your point and I find it a valid comparison in the examples you showed.

I saw your example using the maximum values 1000000000, 10000000. The normalization for the front `{{100000000, 1},{55,55},{1, 100000000}}` would be something like this: `(100000000 - 1)/(100000000 - 1) = 1`; `(1 - 1)/(100000000 - 1) = 0`; `(55 - 1)/(100000000 - 1) = 0.00000054`. Which I think is right: the value 100000000 is the worst possible, thus 1. 55 is very close to 0 (when put to scale), thus 0.00000054. The great HV is due to the point `{55, 55}` which has a low value when normalized. Even though, the HV of the true Pareto would be better than this one. The scale of 0.9999989200002808 is indeed not beautiful, but if the HV values of similar fronts vary in the 6th decimal place, then I don't see the problem. The fact is that, if I have a front composed only by `{{100000000, 1}, {1, 100000000}}`, then its HV would be very close to 0. The difference between 0.9999989200002808 and 0 is compatible with the difference in the fronts (`{55, 55}` is really a good solution).

Aside from that, I still don't think that the way it is being done today is the best way to normalize objective values. I will give you 2 main reasons, which I observed from the struggles I had in the past.

### 1 - Comparing bad fronts

Let me give an example why I think that an HV of 0 for solutions dominated by the reference point is not very useful. Let the true Pareto front be `{{2, 7}, {3, 5}}`, and the results from three algorithms A, B and C be respectively: `{{2, 7}, {3, 5}}`; `{{12, 9}, {11, 10}}`; and `{{15, 13}, {16, 12}}`. Even though the algorithm A had the best values and the true Pareto front is composed solely by its solutions, I still want to compare the algorithms B and C. However, their HV would be both 0, because after the normalization all solutions in their fronts will be dominated by the reference point `{1, 1}`.

This also happens when the true Pareto front is composed by only one solution (which happened in plenty of cases in my experiments). Any approximation set which is covered by the true Pareto front would result in an inaccurate HV, or all the same one: 0.0. Worse than that is when I am trying to calculate the HV average for 30 independent runs, which usually have some bad fronts, and thus the average would be compromised.

However, this is not my main concern. The real problem is the following:

### 2 - The importance of equal scales

One of the main purposes behind normalization and its importance for indicators is that it puts all scales between 0 and 1. This is important because when you have different scales, the indicators may tend to prioritize the objective with the greatest scale. Let me give you a real example. The following front was found after 30 independent runs of an algorithm:

    720.0 0.02040816326530612
    713.0 0.02127659574468085 
    709.0 0.023255813953488372 
    711.0 0.022222222222222223 
    705.0 0.02564102564102564 
    707.0 0.024390243902439025

Note that on objective varies from 720 to 705, while the other varies from 0.020 to 0.025. When I computed the HV (not normalized), not matter how much objective 2 was optimized, HV would always prioritize objective 1 because it decreased more and thus the HV area would have a greater value. After the normalization was done, both objectives were set to the interval of 0 and 1, and the HV started to behave properly. This also happened with other indicators.

Whit the current normalization, the scales are not set equally, as you could see in my examples. Therefore, even if I set a reference point in a way that it is dominated by all solutions, one objective could have the prioritization because the scales are messed.

Juanjdurillo:
Hi,

First of all: nobody said here that the way it is being done is the best
way. Not me at least.

Regarding your first point. I think that in that case you are mentioning,
the problem then is that you are normalizing using the true Pareto front
which is much better than the fronts you are computing. If you know both
fronts are of poor quality, I would use the union of both fronts to compare
between them: we did it in many experiments, specially in those where we do
not know where the true Pareto front is. Notice that this approach do not
require you change the normalization being done, but the input front that
you provide.

Regarding your second point. I do not get it. Nobody is arguing here about
the need to normalize. We just normalized until now using the maximum and
minimum of the true Pareto front (that was what I mentioned when I said we
scale it). I believe, please correct me if I am wrong, that in that example
you mention the normalization the way we are doing also normalizes both
objectives to the 0-1 interval.

2015-05-12 14:55 GMT+02:00 Giovani Guizzo <notifications@github.com>:

GiovaniGuizzo:
Hello Juan.

I know you didn't say it is the best way. I am not arguing that.

I also did what you did in previous experiments. Joined the fronts and informed it as a true Pareto front for obtaining the max and min values. But what I found is that it is not very intuitive, I struggled first to understand what I had to do by analyzing the code. Therefore, in the commit, I created methods for HV, IGD and GD which receive the max and min values for when the user doesn't know that this union must be done for a properly obtainment of max and min values. Nevertheless, the old methods were maintained in the classes.

Ok, I understood your thoughts on the second point. Well, I think I didn't expressed my self correctly then. The problem is that, in the way it is done today, the normalization not always keeps the values between 1 and 0 when using solely the true Pareto front. Let me use again the example in the first comment: true Pareto front: `{{2, 7}, {3, 5}}` and the solution A `{12, 9}`. Using solely the true Pareto front max values (3, 7) and min values (2, 5), the normalization for A would be `{10, 2}`. Of course, if I join the solution A with the true Pareto front, the max would change and the normalization would be done correctly, but it is not very intuitive because the method asks for the true Pareto front not a union of fronts. I didn't say that the normalization must change, but how it is used with the maximum and minimum values by the indicators. But I understand your point and I was concerned that other users might want to use the current normalization, thus I kept the previous methods in the indicator classes.

Juanjdurillo:
In that I can only agree with you: if the input is called Paret front, there is no way the user can reasoning in another way and use a different input. This is definitely a confusing name. It is this kind of things what we are trying to improve in the next version of jMetal.

GiovaniGuizzo:
Exactly. This is in fact the source of the problem. I have several colleagues which used the HV class in a wrong way because of this, or even used WFG instead because in their view "the HV of jMetal is behaving oddly". I also, last year, created an horrendous code to generate a WFG script to compute the HV because it was my view also. A colleague of mine used to emulate the PFtrue by giving as argument to the HV something like this: `{{MAX, MIN, MIN}, {MIN, MAX, MIN}, {MIN, MIN, MAX}}` to force the method to use the real max and min values.

However this problem is not exclusive to the HV class. Take the IGD for instance, where the PFtrue must be the real PFtrue, and not the union of all fronts. If you inform as parameter the union of all fronts, the IGD calculation might be flawed. It seems more intuitive for the method to receive the PFtrue, the approximated PF, the max and the min values. That is why I decided to add the methods which receive the maximum and minimum, and also the calculators which will help the user in doing this union and transparently obtaining the max and min.

Juanjdurillo:
I only partially agree. If you want to compute the quality indicators normalizing using the true Pareto front then I still believe what we have is not wrong ( for the records: not saying is the best ). The fact that we may provide a different API to use another front for normalization does not imply what we are doing is wrong. In my view, again, you see a problem where I am not seeing it: if you know the true Pareto front you should normalize only using its values. Normalizing using other values seems to me a way of justifying results which are far away from the true Pareto front, though justifiable if you actually do not know the true Pareto front.

GiovaniGuizzo:
And I agree with our disagreement :smiley: Even though I think it is not correct, I kept both methods, because, well, this is not my project. It is just a struggle I had and I am concerned that other users may also have.

matthieu-vergne:
So what happened with this pull?

Juanjdurillo:
I have to check it. If both methods are implemented I guess the pull can be accepted

PR #64, opened 2015-05-19
Few improvements on listener time measure

matthieu-vergne:
No functional change, just a fix and stuff which does not change behaviours.


matthieu-vergne:
Sorry for the mess in managing the branch. {'^_^}
I made some cleaning and I went a bit too far.
You can merge it anyway, nothing changes, it is only a management issue from my side.

PR #68, opened 2015-05-19
Generics on quality indicators

matthieu-vergne:
I think that you know there is a lot, lot, lot of work to do regarding the generics, because you let almost all of them empty thus they don't provide anything else than warnings. I have started revising them and I plan to go ahead with that, because we cannot say in the paper that we use generics to have a better control at compilation if it is to use them in such a way that they are useless. So here is an example of refinement and it would be nice that you do it with some others in order to build some habits. Otherwise you will just add warnings after warnings with useless empty generics.

Here is this one:
```
public interface QualityIndicator extends Serializable, DescribedEntity {
  public double execute(Front frontA, Front frontB) ;
  public double execute(List<? extends Solution> frontA, List<? extends Solution> frontB) ;
  public String getName() ;
}
```
First of all, 2 warnings tell you that you did not specify the parameters for `Solution` in the second method (1 warning per front).

Here, my assumption is that you don't care about which kind of `Solution` it is, because this interface is not intended to deal with specific types of data. And here is how it is translated:
```
  public double execute(List<? extends Solution<?>> frontA, List<? extends Solution<?>> frontB) ;
```
The question mark should be used when you **do not care** about the type (not when you don't know). Otherwise, it should be specified with a proper type or a parameter. When you do not specify any parameter by letting `Solution` alone, it is interpreted as `Solution<?>`. And now we will see why it is a problem.

So, what does that mean? It means that you can do all of that:
**SAME SOLUTION CLASS**
```
List<MySolution> frontA = ...;
List<MySolution> frontB = ...;
execute(frontA, frontB);
```
where `MySolution` is for instance a `Solution<String>`, but also:
**SAME SOLUTION PARAMETER**
```
List<Solution<Integer>> frontA = ...;
List<Solution<Integer>> frontB = ...;
execute(frontA, frontB);
```
as well as, by extension:
**SIMILAR SOLUTION CLASS**
```
List<MySolution1> frontA = ...;
List<MySolution2> frontB = ...;
execute(frontA, frontB);
```
where `MySolution1` and `MySolution2` are two different `Solution<Integer>`, but also:
**DIFFERENT SOLUTION PARAMETER**
```
List<Solution<Integer>> frontA = ...;
List<Solution<Double>> frontB = ...;
execute(frontA, frontB);
```
and so of course:
**DIFFERENT SOLUTION CLASS**
```
List<MySolution1> frontA = ...;
List<MySolution3> frontB = ...;
execute(frontA, frontB);
```
where `MySolution1 extends Solution<Integer>` and `MySolution3 extends Solution<Double>`

Now, I assume that you were not expecting the two last cases (maybe three), because you want to have two compatible fronts. If it is not the case, you can close this pull request, because I took the additional assumption that they should be compatible, but please read the remaining to know how a given interpretation is translated into generics.

If you want to forbid the two last cases, then you probably want to have the same parameters, and this come with this definition:
```
  public <T> double execute(List<? extends Solution<T>> frontA, List<? extends Solution<T>> frontB) ;
```
Thus, you are telling that there is a given type `T` which is common to both `Solutions`. In such a case, what you forbid are the cases where solutions are *fundamentally* different. `MySolution1` and `MySolution2` can still be used together, because they both extends `Solution<Integer>`, thus they are still similar.

Now let assume that you expect this `execute()` method to be used on solutions coming from the same algorithm, thus they should be as similar as possible (regarding their class, of course, not their data). Thus, you don't want similar solution classes like `MySolution1` and `MySolution2` to be used together. If you reject this assumption, you can close this pull request too, because I take it.

If you want to restrict to equivalent classes, then you should not say that only the parameters of the solutions are the same, but that the solutions themselves are the same, so you should also factor the part `? extends Solution`, not just the solution parameter:
```
public <T, S extends Solution<T>> double execute(List<S> frontA, List<S> frontB) ;
```
Here we factor the definition by creating `S`, and both lists use it, thus they deal with exactly the same type, so you don't allow anymore to use `MySolution1` and `MySolution2` even if they extend the same `Solution<Integer>`. Now you could say that this case should not hold anymore:
```
List<Solution<Integer>> frontA = ...;
List<Solution<Integer>> frontB = ...;
execute(frontA, frontB);
```
Because it is the same thing. But no, it is not: `MySolution1` and `MySolution2` are *specializations* of `Solution<Integer>`. So what happen concretely? Java takes the most restrictive types and assess their strict equivalence. Here, `MySolution1` is not `MySolution2` so it does not accept it. If you try it with a parent class and a child class, same thing. However, for the case where you provide two `Solution<Integer>`, then the most restrictive it can find is precisely this level. It cannot go further. And at this level, we can only assess the equivalence, thus it accepts it. Notice that you cannot cast the arguments to `List<Solution<Integer>>` in order to make it works, because the casting is effective on the root type (`List`), not on its parameters.

Now, if we look closer the definition:
```
public <T, S extends Solution<T>> double execute(List<S> frontA, List<S> frontB) ;
```
Here, it happens that the parameter of the solution `T` is used only once in `S extends Solution<T>`, thus it does not make any sense anymore to have it and we can rewrite it that way:
```
public <S extends Solution<?>> double execute(List<S> frontA, List<S> frontB) ;
```
We are still defining a single `S` which should be used for both lists, but we don't care about the solution parameter. You can add all the parameters you want, but at the end you can replace all the ones which are used only once by `?` and remove their definition, because what generics allow is to describe type equivalences. So if you use it once, no equivalence (which requires at least two elements) is described, so you can remove it.

This is the version I have used in this commit, and I have adapted the implementation correspondingly. If it is not what you want, feel free to close without merging, but please consider to fill your parameters in the future, otherwise you will just add work for revision, because not filling the generic parameters is like not using parameters at all.

Finally, maybe you noticed that we defined the generics at the level of the method itself, not at the class level (e.g. `QualityIndicator<S extends Solution<?>>`). When you introduce generics, it is better to start from the method. One need to declare a generic at the class level is when you want to have equivalences among different methods of the class. Thus, if you have two methods which should deal with the same type of data, then you will define their generic at the class level instead of the method level, because each method definition is independent. This also implies that it cannot change: by defining it at the method level, the concrete type used is determined at the call of the method. Calling the same method from the same instance with different data is possible, as long as they satisfy the constraints when we call the method. By having the generic at the class level, you fix it when you instantiate the instance, thus any call to the method will deal only with the type specified at the instanciation. This is the second case where you can be interested in defining a generic at the class level: when a given instance should always deal with the same type of data, and not adapt at each call.

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/QualityIndicator.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/QualityIndicator.java
@@ -33,6 +33,6 @@ import java.util.List;
  */
 public interface QualityIndicator extends Serializable, DescribedEntity {
   public double execute(Front frontA, Front frontB) ;
-  public double execute(List<? extends Solution> frontA, List<? extends Solution> frontB) ;
+  public <S extends Solution<?>> double execute(List<S> frontA, List<S> frontB) ;
   public String getName() ;
 }
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/hypervolume/impl/PISAHypervolume.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/hypervolume/impl/PISAHypervolume.java
@@ -40,8 +40,8 @@ return null ;
   }
 
   @Override
-  public double execute(List<? extends Solution> paretoFrontApproximation,
-      List<? extends Solution> trueParetoFront) {
+  public <S extends Solution<?>> double execute(List<S> paretoFrontApproximation,
+      List<S> trueParetoFront) {
 
     if (paretoFrontApproximation == null) {
       throw new JMetalException("The pareto front approximation object is null") ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/hypervolume/impl/WFGHypervolume.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/hypervolume/impl/WFGHypervolume.java
@@ -26,7 +26,7 @@ public class WFGHypervolume extends SimpleDescribedEntity implements Hypervolume
   }
 
   @Override
-  public double execute(List<? extends Solution> frontA, List<? extends Solution> frontB) {
+  public <S extends Solution<?>> double execute(List<S> frontA, List<S> frontB) {
     return 0;
   }
 
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/Epsilon.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/Epsilon.java
@@ -58,8 +58,8 @@ public class Epsilon extends SimpleDescribedEntity implements QualityIndicator {
   }
 
   @Override
-  public double execute(List<? extends Solution> paretoFrontApproximation,
-      List<? extends Solution> trueParetoFront) {
+  public <S extends Solution<?>> double execute(List<S> paretoFrontApproximation,
+      List<S> trueParetoFront) {
 
     if (paretoFrontApproximation == null) {
       throw new JMetalException("The pareto front approximation list is null") ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/ErrorRatio.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/ErrorRatio.java
@@ -49,8 +49,8 @@ public class ErrorRatio extends SimpleDescribedEntity implements QualityIndicato
   }
 
   @Override
-  public double execute(List<? extends Solution> paretoFrontApproximation,
-      List<? extends Solution> trueParetoFront) {
+  public <S extends Solution<?>> double execute(List<S> paretoFrontApproximation,
+      List<S> trueParetoFront) {
 
     if (paretoFrontApproximation == null) {
       throw new JMetalException("The pareto front approximation list is null") ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/GeneralizedSpread.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/GeneralizedSpread.java
@@ -64,8 +64,8 @@ public class GeneralizedSpread extends SimpleDescribedEntity implements QualityI
   }
 
   @Override
-  public double execute(List<? extends Solution> paretoFrontApproximation,
-      List<? extends Solution> trueParetoFront) {
+  public <S extends Solution<?>> double execute(List<S> paretoFrontApproximation,
+      List<S> trueParetoFront) {
 
     if (paretoFrontApproximation == null) {
       throw new JMetalException("The pareto front approximation object is null") ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/GenerationalDistance.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/GenerationalDistance.java
@@ -58,8 +58,8 @@ public class GenerationalDistance extends SimpleDescribedEntity implements Quali
   }
 
   @Override
-  public double execute(List<? extends Solution> paretoFrontApproximation,
-      List<? extends Solution> trueParetoFront) {
+  public <S extends Solution<?>> double execute(List<S> paretoFrontApproximation,
+      List<S> trueParetoFront) {
 
     if (paretoFrontApproximation == null) {
       throw new JMetalException("The pareto front approximation list is null") ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/Hypervolume.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/Hypervolume.java
@@ -57,8 +57,8 @@ public class Hypervolume extends SimpleDescribedEntity implements QualityIndicat
   }
 
   @Override
-  public double execute(List<? extends Solution> paretoFrontApproximation,
-      List<? extends Solution> trueParetoFront) {
+  public <S extends Solution<?>> double execute(List<S> paretoFrontApproximation,
+      List<S> trueParetoFront) {
 
     if (paretoFrontApproximation == null) {
       throw new JMetalException("The pareto front approximation object is null") ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/InvertedGenerationalDistance.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/InvertedGenerationalDistance.java
@@ -61,8 +61,8 @@ public class InvertedGenerationalDistance extends SimpleDescribedEntity implemen
   }
 
   @Override
-  public double execute(List<? extends Solution> paretoFrontApproximation,
-      List<? extends Solution> trueParetoFront) {
+  public <S extends Solution<?>> double execute(List<S> paretoFrontApproximation,
+      List<S> trueParetoFront) {
 
     if (paretoFrontApproximation == null) {
       throw new JMetalException("The pareto front approximation object is null") ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/R2.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/R2.java
@@ -88,8 +88,8 @@ public class R2 extends SimpleDescribedEntity implements QualityIndicator{
   }
 
   @Override
-  public double execute(List<? extends Solution> paretoFrontApproximation,
-      List<? extends Solution> trueParetoFront) {
+  public <S extends Solution<?>> double execute(List<S> paretoFrontApproximation,
+      List<S> trueParetoFront) {
 
     if (paretoFrontApproximation == null) {
       throw new JMetalException("The pareto front approximation object is null") ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/SetCoverage.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/SetCoverage.java
@@ -51,8 +51,8 @@ public class SetCoverage extends SimpleDescribedEntity implements QualityIndicat
   }
 
   @Override
-  public double execute(List<? extends Solution> solutionListA,
-      List<? extends Solution> solutionListB) {
+  public <S extends Solution<?>> double execute(List<S> solutionListA,
+      List<S> solutionListB) {
 
     if (solutionListA == null) {
       throw new JMetalException("The list A is null") ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/Spread.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/Spread.java
@@ -61,8 +61,8 @@ public class Spread extends SimpleDescribedEntity implements QualityIndicator {
 
 
   @Override
-  public double execute(List<? extends Solution> paretoFrontApproximation,
-      List<? extends Solution> trueParetoFront) {
+  public <S extends Solution<?>> double execute(List<S> paretoFrontApproximation,
+      List<S> trueParetoFront) {
 
     if (paretoFrontApproximation == null) {
       throw new JMetalException("The pareto front approximation object is null") ;


ajnebro:
> I think that you know there is a lot, lot, lot of work to do regarding the generics, because you let almost all of them empty thus they don't provide anything else than warnings. 

I'm aware of that. I have been more focused lately in populating the framework with stuff of the older version than paying attention to that issue. 

About the quality indicators, they are intented to compare two list of solutions according to some performance criterion (spread, hypervolume, generational distance, etc), and the comparison takes only into account, in general, the objective values. The comparison is always problem dependant, so it doesn't make sense to compare different subclasses of `Solution`.

Sometimes we are interested in comparing objective values that are stored in files (e..g, the contents of the `FUN` files returned by two or more algorithms). That's the reason of having a `Point` class and a `Front` class. In the former jMetal version we used `Solution`, but I don't like that approach (only the objectives values are needed, and the rest of fields have to be initiliazed to null or similar). Anyway, Juanjo and I discussed about this approach, because `Point` has a semantic of a geometric object instead of a set of objective values, but we didn't find a better alternative.

matthieu-vergne:
> The comparison is always problem dependant, so it doesn't make sense to compare different subclasses of Solution.

So I understand that my assumptions are correct (having `execute(List<S> frontA, List<S> frontB)` with `S extends Solution<?>`). So I guess you can merge it.

> About the quality indicators, they are intented to compare two list of solutions according to some performance criterion (spread, hypervolume, generational distance, etc), and the comparison takes only into account, in general, the objective values.

Well, a quality indicator can be anything, so I can understand it is hard to design. Basically, when I read "quality indicator", I understand it as a method to which you give a solution (or any other item) and which returns a quality value. Thus, more something like that:
```
interface QualityIndicator<Evaluted, Evaluation> {
  public Evaluation evaluate(Evaluted evaluated);
}
```
And here you see that you have two elements: the thing evaluated (e.g. solution) and its evaluation (e.g. objective value). Then you can have different implementations:
- the evaluation should be computed based on properties of the evaluated, which is the most classical way,
- the evaluation is already stored in the evaluated (e.g. it has been computed for another purpose, or in advance to save resources), so you just retrieve it and return it,
- the evaluation is already stored externally (e.g. in a file), so you should retrieve the right value in the right file (depending on the evaluated) and return it,

Of course, you can do different things within a single implementation, searching for instance first a file cache before to compute it from scratch if the cache does not contain it.

If your intention is to *compare*, then you need a quality *comparator*, and here you can base it on the `Comparator` interface, which intends to take two "things" of the same type and return an integer to tell which one is better, with:
- a positive value if the first is better than the second
- a negative value if the second is better than the first
- a higher absolute value for a higher difference between both

Now, you can design a comparator in two ways. One is to restrict to the minimum by evaluating quality values, without knowing anything about what these values refer to:
```
interface QualityComparator<Evaluation> extends Comparator<Evaluation> {
  public int compare(Evaluation e1, Evaluation e2);
}
```
That would be the most straightforward. Another way is to entail more semantic by comparing the evaluated things:
```
interface QualityComparator<Evaluation, Evaluated> extends Comparator<Evaluated> {
  public int compare(Evaluated e1, Evaluated e2);
  public QualityIndicator<Evaluted, Evaluation> getQualityIndicator();
}
```
where basically `compare()` retrieves the quality indicator through the second method, evaluates both the evaluated things with it, and returns the comparison of the evaluations. Anyway, at the end you have to compare the evaluations, so this will remain. Adding the evaluated seems to me unnecessary and add constraints: you could prefer to use different optimized quality indicators to obtain the quality values of each evaluated thing before to compare them, which is not allowed (but in a tricky way) with the extended comparator.

Now, I could understand that you don't like `Comparator` because it returns an integer, prefering a double for continuity. But anyway following the same scheme would probably lead to a better management and, surely, to something more user-friendly because comparators are usual stuff that Java programmers should know about. So you could define it completely from scratch like that:
```
interface QualityComparator<Evaluation> {
  public double compare(Evaluation e1, Evaluation e2);
}
```
Then, you can implement specific quality comparators for lists of elements, but it will return a single double for the whole lists (this interface is for a single comparison, whatever you compare). And it is still straightforward to translate it into a `Comparator` to use sorting methods:
```
Comparator<Evaluation> sortingComparator = new Comparator<Evaluation>() {
  public int compare(Evaluation e1, Evaluation e2) {
    double d = qualityComparator.compare(e1, e2);
    return d > 0 ? 1 : d < 0 ? -1 : 0;
  }
}
```

Going further, if you want to make multiple comparisons then you should go with another interface like this one:
```
interface MultipleQualityComparator<Evaluation> {
  public List<Double> compare(List<EvaluationCouple<Evaluation>> couples);
  
  public static interface EvaluationCouple<Evaluation> {
    public Evaluation getEvaluation1();
    public Evaluation getEvaluation2();
  }
}
```
so for a couple of evaluation at a given index in input you provide a double value at the same index in output.

ajnebro:
> Well, a quality indicator can be anything, so I can understand it is hard to design. Basically, when I read "quality indicator", I understand it as a method to which you give a solution (or any other item) and which returns a quality value. Thus, more something like that:

> ```java 
interface QualityIndicator<Evaluted, Evaluation> {
  public Evaluation evaluate(Evaluted evaluated);
```
The concept of quality indicator in multi-objective optimization is applied only to compare fronts of solutions, which are the results of executing some algorithm. A typical use is when one of the fronts is the true Pareto front (that is usually known in benchmark problems), so the goal is to measure the diversity, the convergence, or both factors of the obtained front having the Pareto front as a reference. 

The quality indicators are typically applied offline, after the execution of the algorithms, but they can be also applied online (e.g., to stop when the current population achieves a particular indicator value when compared with the true Pareto front).

**Comment by matthieu-vergne** [comment]
So somehow, your intepretation of quality indicator is reference-based: you have a given reference (e.g. true pareto front) and the quality indicator tells you how "far" you are from this reference. It is still something like that (ou can simplify by replacing `Evaluation` by `double` from the start):
```
interface QualityIndicator<Evaluted, Evaluation> {
  public Evaluation evaluate(Evaluted evaluated);
}
```
but for a given implementation you will have for instance:
```
class ParetoQualityIndicator<S extends Solution<?>>
                                       implements QualityIndicator<List<S>, Double> {
  private final List<S> reference;

  public ParetoQualityIndicator(List<S> reference) {
    this.reference = reference;
  }

  public Double evaluate(List<S> evaluated) {
    double evaluation;
    // ...
    // compare the evaluated to the reference
    // ...
    return evaluation;
  }
}
```
Something like that would make sense for me, given what you say.

**Comment by ajnebro** [comment]
> So somehow, your intepretation of quality indicator is reference-based: you have a given reference (e.g. true pareto front) and the quality indicator tells you how "far" you are from this reference.

Yes, that's the general case, but you were right when you said that a quality indicator can be anything. Thus, the Set Coverage is an indicator comparing two fronts and none of them is the true Pareto front. 

The result of a comparison can be a distance, a volume, a percentage, etc, so as far as I know a quality indicator always returns a double value.

**Comment by matthieu-vergne** [comment]
For the set coverage, I don't see why you qualify it as a quality indicator if none of the two fronts is a reference. One does not need to be a true front to be a reference, if it is for instance the best you know it is fine to. The point is that the reference does not have the same consideration than the evaluated one. If they are both at the "same level", then whether you are actually making a *comparator*, whether you are computing something else than some quality evaluations.

Could you clarify your use of this set coverage and other similar indicators you have?

**Comment by ajnebro** [comment]
Quality indicators in multi-objective optimization is a complex issue. Many people use the term "metrics", but we prefer "quality indicators" because that's what they are: measures to somehow assess the quality of a front according to certain criteria. As a consequence, you can find two indicators to measure convergence (e.g., generational distance and epsilon) and sometimes obtain contradictory results.

It is not clear about which quality indicators should be used, although there is an agreement in that the hypervolume and the epsilon are very reliable.

You have to consider that when solving real world problems, the Pareto front is not known (otherwise, the problem would be already solved), so the approach to use quality indicators is to build a reference front with all the solutions obtained in all the experiments. Of course, that gives a relative performance measure: if all the algorithms yield poor results, the indicator will allow to say which algorithm is better, not if they are far from the true Pareto front.

Is the set coverage a quality indicator?. If you take a look to a [reference paper](http://e-collection.library.ethz.ch/eserv/eth:25454/eth-25454-01.pdf), the set coverage can be found on table 4. It allows to compare two fronts in terms of convergence, so the answer is yes.

**Comment by matthieu-vergne** [comment]
It seems you are focusing on whether X or Y are relevant quality indicators, but I don't care about X and Y (the instances), I care about what they are (the class). A quality indicator being *relevant* or not is not my issue here. I am speaking about definitions: an indicator takes a single element and gives you a single value about this element, value which represents the level of quality of this element on a particular dimension. If it takes a set of elements, it gives a value related to the full set, not parts of it. In the case where you evaluate an item (like a front) compared to a reference, you are evaluating the item *based on* the reference, not the item *and* the reference together. The reference is an internal property of the indicator, and the item is the sole argument on which you apply the actual evaluation.

This is really a conceptual issue here. Technically, we don't care, any representation is fine and one can implement everything he wants with any interface. But if we speak about something which should be conceptually clear, then we have to provide a representation which highlight the relevant properties so that it is clear what we are doing. When you tell me that many people use the term of "metrics" and you prefer quality indicators "because that's what they are", I feel it quite egocentric. There is definitions for each term. The point is not to speak about personal preferences anymore, but about concrete consensus and well grounded definitions. It can be broadly used definitions, but these definitions should be clear. If it is just "what happens to be used" but no clear definition is provided, then we cannot rely on this term, because people reading it will have different interpretations. And if these interpretations can be somehow compatible at high level, once you have to implement them you have to decide what it really means, and this is where people would disagree.

If you speak about a quality indicator, I expect to see something like `value = indicator.evaluate(item)`. If you speak about a quality comparator, I expect to see something like `comparison = comparator.compare(item1, item2)`. If I see `value = indicator.evaluate(item1, item2)`, that is confusing: am I evaluating item1 based on item2? item2 based on item1? the combination of item1 and item2? Each case can be clarified:
- item1 based on item2:
```
indicator = new Indicator<Item>(item2);
value = indicator.evaluate(item1);
```
- item2 based on item1: 
```
indicator = new Indicator<Item>(item1);
value = indicator.evaluate(item2);
```
- combination of item1 and item2: 
```
indicator = new Indicator<Couple>();
couple = new Couple(item1, item2);
value = indicator.evaluate(couple);
```

So when I read:
> the Set Coverage is an indicator comparing two fronts

For me you it is inconsistent. An indicator is not a comparator. You can have for instance an indicator based on a comparator:
```
comparator = new Comparator<Item>();
comparison = comparator.compare(item1, item2);

reference = item2;
referenceIndicator = new indicator<Item>() {
  public Value evaluate(Item item) {
    return comparator.compare(item, reference);
  }
}
value = referenceIndicator.evaluate(item1);

coupleIndicator = new indicator<Couple>() {
  public Value evaluate(Couple couple) {
    return comparator.compare(couple.leftItem(), couple.righItem());
  }
}
value = coupleIndicator.evaluate(new Couple(item1, item2));
```
What is important here is the interpretation: when you use a comparator, you compare two (or more) items to get a value which relates to both, while when you use an indicator, you evaluate a single element to get a value about this element only. In the absolute, the indicator can compare two elements to return a distance, that is fine, but what is important when you read the code is that you say "I get an evaluation of this item", period. Then, how you get this evaluation is the problem of the implementation of the indicator. A reference would be a parameter of the whole indicator, not an argument of its evaluation function.

> The result of a comparison can be a distance, a volume, a percentage, etc

Not at all:
- a distance is an indicator on a couple which is usually commutative (`d(a,b) = d(b,a)`), while a comparison would tell you which one is more/less distant to a reference, and optionally how much.
- a volume is an indicator on an item, if you want to compare you need to evaluate the volume of another item and compare the volumes.
- a percentage has no unit, so it can be anything.

So what are you using? Indicators or comparators? From my point of view you are using indicators. You can use indicators based on comparators, as well as comparators based on indicators, but they are not equivalent stuff, and a good conceptualization makes it clear what to use for which purpose.

**Comment by ajnebro** [comment]
> When you tell me that many people use the term of "metrics" and you prefer quality indicators "because that's what they are", I feel it quite egocentric

I'm not the one who proposed that "quality indicator" is a more precise term than "metric". Just take a look to the paper I included before. The authors are very relevant in the field and they prefer the former term. I merely agree with them.

**Comment by ajnebro** [comment]
> If you speak about a quality indicator, I expect to see something like `value = indicator.evaluate(item)`. If you speak about a quality comparator, I expect to see something like `comparison = comparator.compare(item1, item2)`. If I see `value = indicator.evaluate(item1, item2)`, that is confusing: am I evaluating item1 based on item2? item2 based on item1? the combination of item1 and item2?

I agree. A quality indicator must return a value and, even when its main use is to compare different fronts, it is not a comparator.

**Comment by ajnebro** [comment]
>So when I read:
>> the Set Coverage is an indicator comparing two fronts

I see now that I gave a very poor definition of the set coverage: " given two nondominated sets, the set coverage calculates the fraction of the second front that is dominated ("covered") by the first one".

**Comment by ajnebro** [comment]
> From my point of view you are using indicators. You can use indicators based on comparators, as well as comparators based on indicators, but they are not equivalent stuff, and a good conceptualization makes it clear what to use for which purpose.

I totally agree.

**Comment by matthieu-vergne** [comment]
Sorry I am a bit harsh {'^_^}, but for me definitions are of first importance because they tell how to design the root interfaces on which all the framework will stand. So I cannot be too open, otherwise we will surely face limitations which require explosive modifications.

> I'm not the one who proposed that "quality indicator" is a more precise term than "metric". Just take a look to the paper I included before.

And if I have well understood this paper, their definition 2 tells that their quality indicator assigns a real value to a vector (`real = indicator.evaluate(vector)`), while their definition 3 tells that the comparison method takes 2 vectors and returns a boolean (`boolean = comparator.compare(vector1, vector2)`). So first, it does not appear as incompatible with what I say. Second, this is strengthen in the text:

> The measures discussed above are examples for unary and binary quality indicators; however, in principle a quality indicator can take an arbitrary number of arguments.

This is what I mean by highlighting the relevant properties: for a quality indicator, even if you are evaluating a set, you don't really care about the size nor the content of this set *at first*. What is important is that you evaluate the full set. They also add:

> Van Veldhuizen and Lamont [22], for instance, applied a combination `I = (Igd,Is,Ionvg)` of three indicators [...]. Accordingly, the combination (or quality indicator vector) `I` can be regarded as a function that assigns each approximation set a triple of real numbers.

Thus, `triple = indicator.evaluate(approximationSet)`. So you can imagine that there is quality indicators which does not return doubles. Yet it is a single value: it is a single point in a 3d space (you could implement it with a `List` or make your own `Triple` class for stronger constraints). Once again, you could say that actually the indicator has many arguments and returns three double values instead of one, but when you consider the definition of this indicator, the focus (what we care about first) is that we evaluate 1 thing (approximation set) and give it 1 value (a triple of reals), and this is what I expect to find in common for any quality indicator, thus motivating my interfaces. While if you think about a comparison definition, you would focus on the multiple arguments, and it would not come to your mind to define a comparison function by giving it all its arguments as a single vector, because the focus is on comparing these arguments, so they have to appear explicitly. Formally, that would be exactly the same, and technically you could obtain same functionalities with same performances. But conceptually, one naturally appears as a better choice because it highlights the relevant properties, while the other highlight something else.

This is what I care about: having a good definition, meaning a good root interface. Then you can implement what you want and combine different stuff in many ways to make highly advanced stuff, but it should start from a well grounded basis which provides all the required flexibility while preserving all the important properties. And for me, an indicator with an evaluation function having two arguments is inadapted. And for the set coverage case, you give the reason:

> given two nondominated sets, the set coverage calculates the fraction of the second front that is dominated ("covered") by the first one

And here I see two interpretations:
- what you care is to assign a quality value to the first, thus the second one is here as a reference (`new Indicator<Item>(second).evaluate(first)`)
- what you care is to assign a quality value to the couple (i.e. ordered pair), thus whatever the first nor the second, we care about the combination (`new Indicator<Couple>().evaluate(new Couple(first, second))`)

By having the right writing, you support the right interpretation.

> I totally agree.

So should I expect that you (*) will introduce something like that in jMetal?
```
interface QualityIndicator<Evaluted, Evaluation> {
  public Evaluation evaluate(Evaluted evaluated);
}
```
or is there something else to discuss about?

(*) I can introduce the interface but I cannot reliably apply it to the jMetal code, because I don't know enough each algorithm to choose between the three translations (A based on B, B based on A, evaluate couple (A,B)).

**Comment by ajnebro** [comment]
Ok, let's go.

If I have undestand everything, we take this interface
```java
interface QualityIndicator<Evaluted, Evaluation> {
  public Evaluation evaluate(Evaluted evaluated);
}
```

and then an indicator such as Epsilon would be defined as:
```java
public class Epsilon extends SimpleDescribedEntity
    implements QualityIndicator<List<Solution<?>>, Double> {

  private List<Solution> referenceFront ;

  public Epsilon(List<Solution> referenceFront) {
    super ("EPSILON", "Additive Epsilon indicator") ;

    this.referenceFront = referenceFront ;
  }

  @Override
  public Double evaluate(List<Solution<?>> solutionList) {
    return execute(solutionList, referenceFront);
  }
  ...
```
and it would be used in this way:
```java
QualityIndicator epsilon = new Epsilon(referenceFront) ;
double value = epsilon.evaluate(front)
```

A positive aspect of this approach is that there is no confusion about the order ot the parameters. In the current approach, with the `execute(Front paretoFrontApproximation, Front trueParetoFront)` method, sometimes it is not clear the parameter order unless you take a look to the documentation.

All the indicators in jMetal can be easily adapted to the new approach. About the set coverage, I'm not sure now whether using:
```java
public class SetCoverage extends SimpleDescribedEntity
    implements QualityIndicator<List<Solution<?>>, Double>{

  private List<Solution<?>> currentFront ;

  public SetCoverage(List<Solution<?>> currentFront) {
    super("SC", "Set coverage") ;

    this.currentFront = currentFront ;
  }

  @Override public Double evaluate(List<Solution<?>> front) {
    return execute(currentFront, front) ;
  }
```
or to take the `Couple` class you mentioned before.

matthieu-vergne:
Given that it is ordered, I would think that it should be a reference-based evaluation.

Trying to look at the uses of this SetCoverage, I did not find any except in the dedicated tests. Do you have any implemented example which uses it? Otherwise, do you have a reference paper which describes it?

ajnebro:
The set coverage (SC) returns a percentage, but SC(A, B) it not necessarily the same of SC(B,A). For this reason, when several fronts are compared, the results are typically arranged in a bi-dimensional table. 

A definition can be found here: http://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=797969

matthieu-vergne:
I would see two things, then.

One is the coverage measure, which takes two vectors of solutions and returns the coverage (of the first over the second one). You could consider it as a comparator, but I would not recommend it: Java comparators are meant for consistent comparisons, meaning [anticommutative](http://en.wikipedia.org/wiki/Anticommutativity)) function (or at least have a consistent sign) with a transitive order. So you should be able to reverse the order by reversing the arguments, which is not the case here (you cannot have negative values).

Another one is the quality indicator, which takes a reference and evaluate a single item. The point of using this indicator is that you have identified a reference, something against which you plan to evaluate all your candidates. If you don't have such a reference, then it could be that you are evaluating the couple, but because this coverage naturally gives a different importance to each argument, then you face the necessity to balance that by evaluating both directions. Thus, the couple-based quality indicator would not return a double, but a couple of doubles ( one for each direction), and your quality evaluation of the couple would be based on that.

ajnebro:
Frequenty, the use of indicators to compare the fronts produced by a number of algorithms, so after computing the indicators the comparisons are carried out by applying statistical tests to ensure that the observed differences are signifcant and then reporting the results in tables containing mean or medians.

In the case of set covegare, as it has to be applied in both senses, if we are going to compare the fronts A, B, C, and D, we have to calculate SC(A,B), SC(A,C) and SC(A,D). From this point of view, perhaps A could be considered as a reference. Later this should be repeated with B, C, and D.

Anyway, I don't dislike the approach of the couple-based quality indicator.

ajnebro:
The remaining issue is the need of using the `Front` class. In most of cases, the reference front is stored in a file which only contains objective values. In the old jMetal, the `QualityIndicator` class has a constructor with a filename as an argument, and it returns a list of solutions. I don't like this idea because these solutions are not true solutions in the sense that they only contain objectives. The approach works because these solutions are only applied to compute the indicator values.

To avoid this, we included a `Front` class which is composed of  `Points`, and that's the reason of two `execute` methods in the old `QualityIndicator` interface (refactoring it to the new one is on my list of next things to do):
```java
public interface QualityIndicator extends Serializable, DescribedEntity {
  public double execute(Front frontA, Front frontB) ;
  public <S extends Solution<?>> double execute(List<S> frontA, List<S> frontB) ;
  public String getName() ;
}
```

This `Front` class is also useful to include some operations (such as invert the front or get a normalized front), which are currently implemented in the `FrontUtils` class ... but I would like to avoid using these kind of put-everything-you-need-in-a-util-class because now there are a `SolutionListUtils`, a `SolutionUtils`, etc. 

Summarizing, perhaps we could leave the `Front` class out and redesign the util classes, but then I'm not sure if we'll ever have a finished version of jMetal 5.0 ;)

matthieu-vergne:
> The remaining issue is the need of using the Front class. In most of cases, the reference front is stored in a file which only contains objective values. In the old jMetal, the QualityIndicator class has a constructor with a filename as an argument, and it returns a list of solutions. I don't like this idea because these solutions are not true solutions in the sense that they only contain objectives. The approach works because these solutions are only applied to compute the indicator values.

I also see another problem here, which is that objective values are computed from the variables of the solution, but here you just retrieve them from a cache, which is not intended by objectives (they are supposed to be computed). Basically, I would expect that your objective values are applied similarly to your "reference front", which is an actual set of solutions. This implies that instead of storing the objective values, you store the variables values. This has 2 advantages:
- you store the original data (the front), so if your objective computation changes at some point, you can just recompute it (while storing the objective values means that your data is now obsolete, but you have no simple way to fix it).
- you actually retrieve proper solutions, so you can apply them without any problem to your indicators and, better, exploit them for instance by displaying the true parete front with the computed one, to visually see how good you are.

So I think that your real issue here is to store computed data instead of raw data. Don't forget that re-computing the true pareto front should be far less consuming than computing hundreds of rounds of populations in algorithms, and once it is computed it remains there as long as you keep the front in memory. You could tell me that you stored the objective values because this is what you are interested about, but this is an optimization: you store the computed value instead of the original one in order to reduce the stored information and computation time, but it has a counter-part and you faced it. Dealing with fake solutions is the cost of storing no information to retrieve them.

On another perspective, you could consider that your quality indicators should take only objective values instead of complete solutions, so it makes sense to store only the objective values (well, it is arguable in general, but for the restricted purpose of comparing objective values only it makes sense). But then, a quality indicator is not about evaluating a solution, but evaluating a set of objective values, so you should provide the facilities to extract the objective values from the solutions. I feel it less natural personnally, but it is another design choice which adds further decoupling but implies additional effort.

Then, anyway, I think that `Front` won't be useful for the quality indicator, in the sense that you can always translate it into a list of solutions and use the list-based `execute()`. Having both of them seems to me irrelevant, unless there is `Front` methods which are specially designed for that, and so it could be actually the reverse: use `Front`, discard the other `execute()`, and add facilities to generate a `Front` from a `List<S>`. But having both of them seems to me clearly irrelevant.

Now, in order to judge about the relevance of keeping `Front`, here is the equivalent I notice with list:
```
public interface Front extends Serializable {
  public void readFrontFromFile(String fileName);
  public int getNumberOfPoints() ; // list.size()
  public int getPointDimensions() ; // already available from the points, so just get one and use point.getNumberOfDimensions(), unless you could need it for empty lists
  public Point getPoint(int index) ; // list.get(index)
  public void setPoint(int index, Point point) ; // list.set(index, item)
  public void sort(Comparator<Point> comparator) ; // Collections.sort(list, comparator)
}

```
For the file reading, I would see it as a constructor (or into a factory), not a method that you can call at any time, but look at (*) for further details. Similarly for `Point` compared to `Solution`:
```
public interface Point {
  public int getNumberOfDimensions(); // solution.getNumberOfObjectives()
  public double[] getValues() ; // you can add it to objectives if you want, although I would use a List<Double> instead of an array, but is it required?
  public double getDimensionValue(int index) ; // solution.getObjective(index)
  public void setDimensionValue(int index, double value) ; // solution.setObjective(index, value)
}
```
so surely I see a lot of redundancy and not much added value, so I could understand that `Front` and `Point` are not needed.

(*) I have this perspective: having static structures is better than having dynamic ones. The inconvenient is obvious: when you want to change it, you need to create a new instance, and if the parameters are provided in a progressive way, then you should use a builder to get the instance at the end instead of just using the right setter at the right time. But on the other hand, the management is simplified because you get all the parameters at the instantiation and then it won't change anymore, so no need to deal with side effects due to dynamicity, and any other entity having an instance won't be tricked (by having a dynamic instance which represents something, one can be tricked because at some point it represents something else without knowing it). Usually, dynamicity is not what I expect from a solution: a solution is a solution, and another solution is another one. Changing a solution does not make much sense, while getting a new one from an existing one makes sense. Then, more broadly, any specific way to create a front can go into a factory (`factory.getFromFile(file)`, `factory.inverse(front)`, `factory.normalize(front)`, etc.). If you only have creation methods, that would be fine. If you have additional methods, I would be curious about them, but I did not find any relevant one by looking at `Front` and its extensions.

You could speak about the mutators and other transformation operators which change solutions, but usually you (A.i) clone the original and (A.ii) change the clone, while you could instead (B.i) get the variable values of the original, (B.ii) change these values, and (B.iii) use them as parameters for creating a new solution. Thus, your operator could focus on the variables values (making it also independent of the solution, because you only get the relevant variables to mutate), thus the step (B.ii), instead of (A.ii) or (A.i)+(A.ii).

> Summarizing, perhaps we could leave the Front class out and redesign the util classes, but then I'm not sure if we'll ever have a finished version of jMetal 5.0 ;)

You can make `Front`, `Point`, and the corresponding methods -including the execute method- deprecated (display warnings when it is used), and deal with them progressively until there is nothing anymore related to them, so that we can delete them. You already have many warnings for generics which seem to me more important, because we speak about one of the paper's arguments. Then decoupling with deprecated stuff can be dealt with progressively. You will probably go through some of them during other improvements too, and I can help if it is clear how to replace a given use of a `Front` by a given use of a `List` of solutions.

ajnebro:
I'm trying to close this issue during the GECCO conference, but again I'm having problems with generics.

The `QualityIndicator` interface is:
```java
public interface QualityIndicator<Evaluate, Result> extends DescribedEntity {
  public Result evaluate(Evaluate evaluate) ;
  public String getName() ;
}
```
and the indicators implementation follow the same implementation scheme. For example, this is part of the `Hypervolume` class:

``` java
public class Hypervolume extends SimpleDescribedEntity implements QualityIndicator<List<Solution<?>>, Double> {
  private Front referenceParetoFront ;

  public Hypervolume() {
    super("HV", "Hypervolume quality indicator") ;
  }

  /**
   * Constructor
   * @param referenceParetoFrontFile
   * @throws FileNotFoundException
   */
  public Hypervolume(String referenceParetoFrontFile) throws FileNotFoundException {
    this() ;
    Front front = new ArrayFront(referenceParetoFrontFile);
    referenceParetoFront = front ;
  }

  /**
   *
   * @param referenceParetoFront
   * @throws FileNotFoundException
   */
  public Hypervolume(Front referenceParetoFront) {
    this() ;
    if (referenceParetoFront == null) {
      throw new JMetalException("The pareto front is null");
    }

    this.referenceParetoFront = referenceParetoFront ;
  }

  @Override public Double evaluate(List<Solution<?>> paretoFrontApproximation) {
    if (paretoFrontApproximation == null) {
      throw new JMetalException("The pareto front approximation is null") ;
    }

    return hypervolume(new ArrayFront(paretoFrontApproximation), referenceParetoFront);
  }
...
}
```

I've used this indicator to evaluate a `List<DoubleSolucion>` that is returned by an algorithm. Then, there are at last three ways of using the indicator:

### Way 1
```java
Front referenceFront = new ArrayFront("jmetal-problem/src/test/resources/pareto_fronts/ZDT1.pf") ;
List<DoubleSolution> population ;

QualityIndicator hypervolume = new Hypervolume(referenceFront) ;
Double hvValue = (Double) hypervolume.evaluate(population);
```
This a logical and compact way of using the indicator, but a casting to Double is needed when calling `evaluate()`

### Way 2
```java
Front referenceFront = new ArrayFront("jmetal-problem/src/test/resources/pareto_fronts/ZDT1.pf") ;
List<DoubleSolution> population ;

QualityIndicator<List<Solution<?>>, Double> hypervolume = new Hypervolume(referenceFront) ;
Double hvValue = hypervolume.evaluate(population);
```

This way there is no need of casting but then we get a compilation error (`evaluate()` is expecting a `List<Solution<?>>` and it doesn't like a `List<DoubleSolution>`), not to mention that creating the hypervolume this way can freighten many people.

### Way 3
```java
Front referenceFront = new ArrayFront("jmetal-problem/src/test/resources/pareto_fronts/ZDT1.pf") ;
List<DoubleSolution> population ;
Hypervolume hypervolume = new Hypervolume(referenceFront) ;

Double hvValue = hypervolume.evaluate(population);
```

This is a logical way of declaring the hypervolume object, its compact, there is no need of casting to Double, but it produces the same compilation error as Way 2.

matthieu-vergne:
The good way is the third one, and the compilation error is due to you definition of the hyper volume: it implements QualityIndicator<List<Solution<?>>, Double>, so it asks a List<Solution<?>> as an argument, which means a list of generic solutions. If you gives it a List<DoubleSolution>, then it is a list of specific solutions and so refuse it. What you need to use in you definition is List<? extends Solution<?>>, which means a list of whatever can be casted to a generic solution (so a generic solution or any extension). The DoubleSolution being an extension, you should not have the issue anymore.

This kind of issue (choosing when to use "? extends xxx" instead of "xxx") is the hardest thing I know with generics, but it is often a definition issue, not a usage one, so the hard stuff is for the one making the lib, not the user.

The second way you show is the right way for generic coding, when you don't care about which exact indicator you use as long as it takes a specific type of argument and returns a specific type of result. If you solve the issue above it should work as well as the thir way. It is more verbose but it allows you to use any other indicator having the same definition.

the first way is the usual mistake: you forget to precise the generics, so it does not know which type of argument and returned value is used. So it assumes the most generic ones (Object) and when you want to store the result into a double, you need to cast.

Le 14 juillet 2015 09:58:52 CEST, "Antonio J. Nebro" <notifications@github.com> a écrit :
>I'm trying to close this issue during the GECCO conference, but again
>I'm having problems with generics.
>
>The `QualityIndicator` interface is:
>```java
>public interface QualityIndicator<Evaluate, Result> extends
>DescribedEntity {
>  public Result evaluate(Evaluate evaluate) ;
>  public String getName() ;
>}
>```
>and the indicators implementation follow the same implementation
>scheme. For example, this is part of the `Hypervolume` class:
>
>``` java
>public class Hypervolume extends SimpleDescribedEntity implements
>QualityIndicator<List<Solution<?>>, Double> {
>  private Front referenceParetoFront ;
>
>  public Hypervolume() {
>    super("HV", "Hypervolume quality indicator") ;
>  }
>
>  /**
>   * Constructor
>   * @param referenceParetoFrontFile
>   * @throws FileNotFoundException
>   */
>public Hypervolume(String referenceParetoFrontFile) throws
>FileNotFoundException {
>    this() ;
>    Front front = new ArrayFront(referenceParetoFrontFile);
>    referenceParetoFront = front ;
>  }
>
>  /**
>   *
>   * @param referenceParetoFront
>   * @throws FileNotFoundException
>   */
>  public Hypervolume(Front referenceParetoFront) {
>    this() ;
>    if (referenceParetoFront == null) {
>      throw new JMetalException("The pareto front is null");
>    }
>
>    this.referenceParetoFront = referenceParetoFront ;
>  }
>
>@Override public Double evaluate(List<Solution<?>>
>paretoFrontApproximation) {
>    if (paretoFrontApproximation == null) {
>  throw new JMetalException("The pareto front approximation is null") ;
>    }
>
>return hypervolume(new ArrayFront(paretoFrontApproximation),
>referenceParetoFront);
>  }
>...
>}
>```
>
>I've used this indicator to evaluate a `List<DoubleSolucion>` that is
>returned by an algorithm. Then, there are at last three ways of using
>the indicator:
>
>### Way 1
>```java
>Front referenceFront = new
>ArrayFront("jmetal-problem/src/test/resources/pareto_fronts/ZDT1.pf") ;
>List<DoubleSolution> population ;
>
>QualityIndicator hypervolume = new Hypervolume(referenceFront) ;
>Double hvValue = (Double) hypervolume.evaluate(population);
>```
>This a logical and compact way of using the indicator, but a casting to
>Double is needed when calling `evaluate()`
>
>### Way 2
>```java
>Front referenceFront = new
>ArrayFront("jmetal-problem/src/test/resources/pareto_fronts/ZDT1.pf") ;
>List<DoubleSolution> population ;
>
>QualityIndicator<List<Solution<?>>, Double> hypervolume = new
>Hypervolume(referenceFront) ;
>Double hvValue = hypervolume.evaluate(population);
>```
>
>This way there is no need of casting but then we get a compilation
>error (`evaluate()` is expecting a `List<Solution<?>>` and it doesn't
>like a `List<DoubleSolution>`), not to mention that creating the
>hypervolume this way can freighten many people.
>
>### Way 3
>```java
>Front referenceFront = new
>ArrayFront("jmetal-problem/src/test/resources/pareto_fronts/ZDT1.pf") ;
>List<DoubleSolution> population ;
>Hypervolume hypervolume = new Hypervolume(referenceFront) ;
>
>Double hvValue = hypervolume.evaluate(population);
>```
>
>This is a logical way of declaring the hypervolume object, its compact,
>there is no need of casting to Double, but it produces the same
>compilation error as Way 2.
>
>
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jMetal/jMetal/pull/68#issuecomment-121157404

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

ajnebro:
They are working now :)

The remaining indicator is set coverage. It is defined this way:

```java
import org.apache.commons.lang3.tuple.ImmutablePair;
import org.apache.commons.lang3.tuple.Pair;

public class SetCoverage extends SimpleDescribedEntity implements
    QualityIndicator<Pair<List<Solution<?>>, List<Solution<?>>>, Pair<Double, Double>> {

  public SetCoverage() {
    super("SC", "Set coverage") ;
  }

  private Comparator<Solution<?>> dominance;

  @Override
  public Pair<Double, Double> evaluate(Pair<List<Solution<?>>, List<Solution<?>>> pairOfSolutionLists) {
    List<Solution<?>> front1 = pairOfSolutionLists.getLeft() ;
    List<Solution<?>> front2 = pairOfSolutionLists.getRight() ;
    Pair<Double, Double> result = new ImmutablePair<>(setCoverage(front1, front2), setCoverage(front2, front1)) ;

    return result ;
  }
...
}
```

and it can be used this way:

```java
    Pair<Double, Double> setCoverageValues =
        (Pair<Double, Double>) setCoverage.evaluate(new ImmutablePair(population1, population2));

    JMetalLogger.logger.info("C(pop1, pop2) : " + setCoverageValues.getLeft());
    JMetalLogger.logger.info("C(pop2, pop1) : " + setCoverageValues.getRight());
```

It is working properly, but I don't know whether it is perhaps a bit complicated to use.

matthieu-vergne:
I notice that you still have the casting of the result, so their is an issue here. Maybe you can fix it now considering the previous case.

I think that your lists should have the same type of solutions, because it does not mean anything to me to check a set coverage between a List<DoubleSolution> and a List<BinarySolution>. Am I correct?

ajnebro:
> I think that your lists should have the same type of solutions, because it does not mean anything to me to check a set coverage between a List<DoubleSolution> and a List<BinarySolution>. Am I correct?

Well, there exits the unlike event of someone solving the same problem by using different encodings, so that kind of checking could happen.

ajnebro:
In GECCO 2015 a new quality indicator was presented, the inverted generational distance plus (IGD+). I'm working on including it.

This is against the idea of closing this issue ASAP, but after being in the presentation of the paper about that indicator I couldn't resists the temptation of implementing it ...

matthieu-vergne:
{^o^} Lol! Hard to resist, uh? Well, we will say that it gives an additional case study. {^_°}

Coming back to your concern, first you should be able to remove the casting `(Pair<Double, Double>)`. Considering that you are in a context where you know that you are using the `SetCoverage` quality indicator, then you should have all the information available, so no casting should be required.

Second, regarding the complexity, you can simplify: you can add a method `evaluate(List<Solution<?> list1, List<Solution<?> list2)`. It can seems contradicting to you, because we made the effort to have a generic, single argument method, but this generic method is for generic purpose. In a context where you know exactly which specific measure you are using, you can (and should) use any feature it provides you to improve your code. The whole point of our generic method is to be sure that we can still use it in a generic context, so where we have only few information about the data at hand. Typically, in such a case, you would have the quality indicator without knowing which kind of indicator it is, but also the pair of lists without even knowing that it is a pair of lists (let's call it "thing"), and all what you would know would be that you can apply your "thing" to your quality indicator in order to obtain a "result" that you can returns to some interested tier. This is the kind of situation where our single-argument method makes the most sense. Don't try to limit yourself to have only teh generic stuff, that would be a bad idea. The important is to have it, not to reduce to it. It is just that if it is well designed then you will often be able to reduce to it, but it should not be a constraint.

Third, I am expecting problems because you ask for `List<Solution<?>` and not `List<? extends Solution<?>`. This is exactly the same problem than the one you mentionned just before.

ajnebro:
> {^o^} Lol! Hard to resist, uh? Well, we will say that it gives an additional case study. {^_°}

You cannot imagine how much ;)

Ten papers presented at GECCO have used jMetal, and I have had discussions with some of the authors of them. Other colleagues are interested in porting their algorithms to the new jMetal, and some interesting ideas have been presented ... the only limitation to not start working with them right now is time.

matthieu-vergne:
Please share the ideas in dedicated issues. It makes a good todolist and if someone can help, it will save you time.

ajnebro:
Well, I have just finished a first version of the `org.uma.jmetal.qualityIndicator.RunIndicator` class: https://github.com/jMetal/jMetal/blob/master/jmetal-exec/src/main/java/org/uma/jmetal/qualityIndicator/RunIndicator.java

To have the chance of having a list of indicators I have refactored all of them to use them in this way:
```java

 Hypervolume<List<DoubleSolution>> hypervolume = new Hypervolume(referenceFront);
```

To show an example of how to use them I've used the `NSGAIIRunner` class: https://github.com/jMetal/jMetal/blob/master/jmetal-exec/src/main/java/org/uma/jmetal/runner/multiobjective/NSGAIIRunner.java
(lines 115 -170).

Now is time to review and to think in making some refactoring.

ajnebro:
I'm having problems when running https://github.com/jMetal/jMetal/blob/master/jmetal-exec/src/main/java/org/uma/jmetal/qualityIndicator/RunIndicator.java from the command line.

After setting the CLASSPATH:
```
export export CLASSPATH=jmetal-core/target/classes/:jmetal-exec/target/classes/:jmetal-problem/target/classes/:jmetal-problem/target/classes/
```
I was using this command:
```
java org.uma.jmetal.qualityIndicator.RunIndicator ALL jmetal-core/src/test/resources/pareto_fronts/Golinski.pf FUN.tsv FALSE
```

but after adding lines 118-124 to use SetCoverage in https://github.com/jMetal/jMetal/blob/master/jmetal-exec/src/main/java/org/uma/jmetal/qualityIndicator/RunIndicator.java I got this:

```
MacBook-Pro-de-Antonio:jMetal ajnebro$ java org.uma.jmetal.qualityIndicator.RunIndicator ALL jmetal-core/src/test/resources/pareto_fronts/Golinski.pf FUN.tsv FALSE
EP: -694.7098067350285
HV: 0.659462954278037
GD: 0.1525979429923311
IGD: 211.88144710386987
IGD+: 0.0
SPREAD: 0.9997882481117343
GSPREAD: 0.9998343627364447
R2: 0.7836157185486684
ER: 1.0
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/commons/lang3/tuple/Pair
	at org.uma.jmetal.qualityIndicator.RunIndicator.calculateAndPrintIndicators(RunIndicator.java:118)
	at org.uma.jmetal.qualityIndicator.RunIndicator.main(RunIndicator.java:44)
Caused by: java.lang.ClassNotFoundException: org.apache.commons.lang3.tuple.Pair
	at java.net.URLClassLoader$1.run(URLClassLoader.java:372)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:360)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 2 more
```
Then I took the alternative way of executing using `mvn`:
```
mvn exec:java -Dexec.mainClass="org.uma.jmetal.qualityIndicator.RunIndicator"
```

without good results:
```
[INFO] Scanning for projects...
[INFO] Inspecting build with total of 5 modules...
[INFO] Installing Nexus Staging features:
[INFO]   ... total of 5 executions of maven-deploy-plugin replaced with nexus-staging-maven-plugin
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO] 
[INFO] org.uma.jmetal:jmetal
[INFO] org.uma.jmetal:jmetal-core
[INFO] org.uma.jmetal:jmetal-problem
[INFO] org.uma.jmetal:jmetal-algorithm
[INFO] org.uma.jmetal:jmetal-exec
[INFO]                                                                         
[INFO] ------------------------------------------------------------------------
[INFO] Building org.uma.jmetal:jmetal 5.0-Beta-35-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- exec-maven-plugin:1.4.0:java (default-cli) @ jmetal ---
[WARNING] 
java.lang.ClassNotFoundException: org.uma.jmetal.qualityIndicator.RunIndicator
	at java.net.URLClassLoader$1.run(URLClassLoader.java:372)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:360)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at org.codehaus.mojo.exec.ExecJavaMojo$1.run(ExecJavaMojo.java:281)
	at java.lang.Thread.run(Thread.java:745)
```

Probably I'm missing some silly detail, but after many hours working on this is time to rest for a while ...

matthieu-vergne:
I start by mentionning that I have lost Internet at my home yesterday (changing contract) and it should come back during the next week (Yes, a really bad timing). For now, I can only read my e-mails when I am at work, so I will probably not be as reactive as usual (especially during this weekend).

I did not look for your inability to run your code, which could take some time if it is tricky (I am used to run jars, not class files which need to set the classpath and so on), so I let you deal with that once you have rest and I will focus on the reviewing. However, maybe you should consider a dedicated module and use the JAR of this module, because people will basically get and use the JARs, not the class files of jMetal (only if they take the sources and compile them, which is *not* the aim of having jMetal on Maven).

# Class Naming

My first comment start from the Javadoc of your `RunIndicator`:
> Class for executing quality indicators from the command line

So I would rename the class as `CommandLineIndicatorRunner` or something similar, because `RunIndicator` means, at least for my English, an indicator about a run, which is not what it seems to be here.

Looking further, you could probably have several `IndicatorRunner` classes depending on where you want to display, or even better: a single one which runs and provides notifications to registered observers, and dedicated displayers which can register observers to get the notifications and display them where they have to (standard output, logger, GUI, etc.). This is the kind of architecture I have used in a recent lib to [manage task progress](https://github.com/matthieu-vergne/Progress): I have the `Progress` object which only manages the progress, and a `Displayer` which can register to this progress, to know when it evolves and display its current status. I have a displayer factory which provides different *immediate* displayers (display each time the progress is updated) as well as *regular* displayers (display status every X seconds). You can even register several displayers to display it on the standard output, the logger, etc. all at once, and all with different strategies (immediate for logger, regular for the output, etc.). Then, you can have a dedicated class like your current `RunIndicator` which would use the runner + the right displayer in a main method.

Refactoring this way does not seem a priority to me, but it would be interesting to go for it if you plan to make several versions of that code soon. The next sections seem to me more important.

# Evaluation-Normalization Decoupling

I have a concern about the normalization process and the `NormalizableQualityIndicator` class, which only manages a normalize flag and let the actual normalization to child classes. If I look for instance the [`Hypervolume` class](https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/impl/Hypervolume.java) (and some others), it seems clear to me that this part should be part of the parent class:
```
double[] maximumValue;
double[] minimumValue;

// STEP 1. Obtain the maximum and minimum values of the Pareto front
maximumValue = FrontUtils.getMaximumValues(trueParetoFront);
minimumValue = FrontUtils.getMinimumValues(trueParetoFront);

// STEP 2. Get the normalized front and true Pareto fronts
normalizedFront = FrontUtils.getNormalizedFront(front, maximumValue, minimumValue);
```
In particular, I would refactor `NormalizableQualityIndicator` to take the reference front in its constructor, so any extension receives it and provide it to the `super()` method, and forget about it. Then, `NormalizableQualityIndicator` should also implement the `evaluate(...)` method, which applies the code above if required and then delegate the remaining computation to an additional abstract method `evaluatePreparedFront(...)` (implemented by the child classes) which takes the (non-)normalized front and compute it, without caring about the normalization stuff.

But personally, I think that the initial design is poor. In particular, you said:
> To have the chance of having a list of indicators I have refactored all of them to use them in this way

When you have such a big change to spread, you should be lazy enough to wonder whether it is worth doing all these changes or not. Personally, I would go for another design: normalization is not reserved to quality indicators, so better to have a `FrontNormalizer` which only do that and do it well, and let the quality indicators focus on their evaluations, and only that (separate the responsibilities). In this `FrontNormalizer`, you would have a constructor taking the reference front, and you would provide a method `normalize(...)` which takes the front to normalize and returns the normalized front. The result would be that all your quality indicators are completely free of any reference front (so they can directly implement `QualityIndicator` and you can remove `NormalizableQualityIndicator`), so they can focus on the evaluation of the front, whatever it is. Here would be the use:
```
List<DoubleSolution> normalizedPopulation
		= new FrontNormalizer(referenceFront).normalize(population);

Hypervolume<List<DoubleSolution>> hypervolume = new Hypervolume()
JMetalLogger.logger.info("Hypervolume (N) : " + hypervolume.evaluate(normalizedPopulation));
JMetalLogger.logger.info("Hypervolume     : " + hypervolume.evaluate(population));
```
I don't know if you are interested in such a decoupling. If you think it is worth doing it, you can continue your reading. If you think it is *not* worth, then I recommend you to stop *immediately* your reading and think deeply about why it is not. Because I believe that once you will read the rest of this post (in particular the next section), it will be really hard for you to remain on that position. {^_^}

Once you have decoupled the normalization from the evaluation, you can remove this crazy stuff:
```
Double hvValueNormalized = hypervolume.evaluate(population);
Double hvValue = hypervolume.setNormalize(false).evaluate(population);
Double epsilonValue = epsilon.evaluate(population);
Double epsilonValueNormalized = epsilon.setNormalize(true).evaluate(population);
// ...
```
which is biohazard code which sometimes set the normalization flag, sometimes not, sometimes to true, sometimes to false. Added to the idea of having a `setNormalize()` which returns the measure to immediately call `evaluate()`, it is just too confusing. Dedicate the normalization to a specific class, and here would be the result:
```
List<DoubleSolution> normalizedPopulation
		= new FrontNormalizer(referenceFront).normalize(population);
Double hvValue = hypervolume.evaluate(population);
Double hvValueNormalized = hypervolume.evaluate(normalizedPopulation);
Double epsilonValue = epsilon.evaluate(population);
Double epsilonValueNormalized = epsilon.evaluate(normalizedPopulation);
// ...
```
that you can further reorganize as:
```
Double hvValue = hypervolume.evaluate(population);
Double epsilonValue = epsilon.evaluate(population);
// and other non-normalized measures

List<DoubleSolution> normalizedPopulation
		= new FrontNormalizer(referenceFront).normalize(population);
Double hvValueNormalized = hypervolume.evaluate(normalizedPopulation);
Double epsilonValueNormalized = epsilon.evaluate(normalizedPopulation);
// and other normalized measures
```
but it is even better if you remove these variables to directly put the calls within the display, because you never reuse the variables anyway:
```
Hypervolume<List<DoubleSolution>> hypervolume = new Hypervolume()

List<DoubleSolution> normalizedPopulation
		= new FrontNormalizer(referenceFront).normalize(population);

JMetalLogger.logger.info("Hypervolume (N) : " + hypervolume.evaluate(normalizedPopulation));
JMetalLogger.logger.info("Hypervolume     : " + hypervolume.evaluate(population));
```
This way, you significantly reduce the amount of code and gain in clarity.

By the way, computing everything before to do all the display, or declaring all your variables at the beginning but assigning/reading them only way later, is a bad habit you (still) have. It makes sense *sometimes*, but only because you need it to achieve what you want. I dedicate another section to that.

# Listing Indicators

In `RunIndicator`, you have this:
```
private static List<QualityIndicator<List<DoubleSolution>, Double>> getAvailableIndicators(
		Front referenceFront, boolean normalize) throws FileNotFoundException {

	List<QualityIndicator<List<DoubleSolution>, Double>> list = new ArrayList<>() ;
	list.add(new Epsilon<List<DoubleSolution>>(referenceFront).setNormalize(normalize)) ;
	list.add(new Hypervolume<List<DoubleSolution>>(referenceFront).setNormalize(normalize)) ;
	list.add(new GenerationalDistance<List<DoubleSolution>>(referenceFront).setNormalize(normalize)) ;
	list.add(new InvertedGenerationalDistance<List<DoubleSolution>>(referenceFront).setNormalize(normalize)) ;
	list.add(new InvertedGenerationalDistancePlus<List<DoubleSolution>>(referenceFront).setNormalize(normalize)) ;
	list.add(new Spread<List<DoubleSolution>>(referenceFront).setNormalize(normalize)) ;
	list.add(new GeneralizedSpread<List<DoubleSolution>>(referenceFront).setNormalize(normalize)) ;
	list.add(new R2<List<DoubleSolution>>(referenceFront).setNormalize(normalize)) ;
	list.add(new ErrorRatio<List<DoubleSolution>>(referenceFront).setNormalize(normalize)) ;

	return list ;
}
```

I think that you already know that it is a really bad way to list available implementations, in the sense that you have to change the code each time a new indicator is provided. Noticeably, having a factory would centralize better the code, but not solve this problem. Here you should exploit the genericity that we have set up with our interfaces, but you can do it easily only if you apply the decoupling I described in the previous section (by using a dedicated `FrontNormalizer`). Because if you do, then all your quality indicators directly implement `QualityIndicator` and do not need anything for their constructor. If you do that, this method would reduce to something like this:
```
private static List<QualityIndicator<List<DoubleSolution>, Double>> getAvailableIndicators() throws FileNotFoundException {

	List<QualityIndicator<List<DoubleSolution>, Double>> list = new ArrayList<>() ;
	list.add(new Epsilon<List<DoubleSolution>>()) ;
	list.add(new Hypervolume<List<DoubleSolution>>()) ;
	list.add(new GenerationalDistance<List<DoubleSolution>>()) ;
	list.add(new InvertedGenerationalDistance<List<DoubleSolution>>()) ;
	list.add(new InvertedGenerationalDistancePlus<List<DoubleSolution>>()) ;
	list.add(new Spread<List<DoubleSolution>>()) ;
	list.add(new GeneralizedSpread<List<DoubleSolution>>()) ;
	list.add(new R2<List<DoubleSolution>>()) ;
	list.add(new ErrorRatio<List<DoubleSolution>>()) ;

	return list ;
}
```
Notice the normalization which does not need to be managed here anymore. You would directly manage it in `calculateAndPrintIndicators()` by replacing:
```
Front front = new ArrayFront(args[2]);
```
by:
```
Front front = new ArrayFront(args[2]);
if(normalize) {
	front = new FrontNormalizer(referenceFront).normalize(front);
} else {
	// keep original front
}
```
or maybe by storing the normalizer in some member variable for reuse.

Once this listing method is simplified, then all your indicators have common points:
1. they are all `QualityIndicator` implementations (they already are)
2. they have all empty constructors (currently not, and it makes the following *a lot* harder, if possible at all)
3. they are all implemented in the `org.uma.jmetal.qualityindicator.impl` package (not a requirements but simplify things too)

If you give the possibility to load new, custom indicators, only the condition 3 would change, but it can be managed quite easily: whether you have a conventional directory in which to put the extensions to jMetal, whether the user explicitely asked to the framework to load a specific file, so you always know where to find the classes and the condition 3 can be adapted consequently. I don't go further in details now for extensions, but this is where you see all the power of genericity. And it starts with your internal indicators: to be able to retrieve them, you need to know which classes to load and to load them through a `ClassLoader`. I don't give further details now, but you can easily find tutorials on the Web, like [this one](http://tutorials.jenkov.com/java-reflection/dynamic-class-loading-reloading.html) for loading a well identified class. You can do this by yourself if you want, but you will still need to provide the paths of each class, so it just makes it worse. To really simplify everything, you need to be able to find which classes are available, what I did for another application (also to manage external JARs and class directories), so I can share it later if you want. You can find tutorials too, but it becomes quite technical.

If you first refine your classes to make the indicators completely decoupled from the normalization process, then I can implement a complete loader if you want, which manages external JARs and directories of classes, with tests and all. I could also do the class refactoring if you are interested, but without Internet at my place, you will have to wait for Monday for the commits, so there would be surely many conflicts to merge. So I prefer that you refactor them during the weekend (if you are interested) and then I can provide you a class loader to reduce your listing method to a mere `QualityIndicatorFactory.getAvailableIndicators()` which does all the magics. With that, we really start to use the genericity features. {^_^}

Additionally, it seems to me weird to specify an argument for the reference front and another for the normalization flag (true/false): if you give a reference front, then normalize, otherwise don't do it. That would seem more logical to me (unless you use your reference front for something else, I did not check).

# Discard C habits

This C habit *must* be fixed if you want to have a clear and naturally optimized code. Taken from `Epsilon.epsilon()`:
```
private double epsilon(Front front, Front referenceFront) throws JMetalException {
	Front normalizedFront;
	Front normalizedParetoFront;

	int i, j, k;
	double eps, epsJ = 0.0, epsK = 0.0, epsTemp;

	int numberOfObjectives = front.getPointDimensions() ;

	eps = Double.MIN_VALUE;

	if (normalize) {
		// No use af any of the local variables declared before
		// Only fill normalizedFront and normalizedParetoFront
	} else {
		// No use af any of the local variables declared before
		// Only fill normalizedFront and normalizedParetoFront
	}

	// eps and numberOfObjectives started to be used only from here
	for (i = 0; i < normalizedParetoFront.getNumberOfPoints(); i++) {
		// epsJ and i used only in this scope
		for (j = 0; j < normalizedFront.getNumberOfPoints(); j++) {
			// epsK and j used only in this scope
			for (k = 0; k < numberOfObjectives; k++) {
				// epsTemp and k used only in this scope
			}
			// ...
		}
		// ...
	}
	return eps;
}
```
Stop declaring everything at the beginning. You force the JVM to maintain *all* these resources during *all* the run of the method. If it is for a value that you store to not recall it later, like `numberOfObjectives`, then it makes sense, but don't store it before you actually need it (just before the for loop, not from the very beginning) and make it `final` to avoid mistakenly rewritting it. However, for variables which are declared way before they are assigned a value, and even declared from a more global scope than they need, it makes no sense at all in Java. We are not optimizing register booking here, you deal with JVM resources, not CPU resources. This:
```
int i, j, k;

// lot of unrelated code

for (i = 0; i < normalizedParetoFront.getNumberOfPoints(); i++) {
	for (j = 0; j < normalizedFront.getNumberOfPoints(); j++) {
		for (k = 0; k < numberOfObjectives; k++) {
			// ...
		}
		// ...
	}
	// ...
}
```
should be replaced by this:
```
// lot of unrelated code

for (int i = 0; i < normalizedParetoFront.getNumberOfPoints(); i++) {
	for (int j = 0; j < normalizedFront.getNumberOfPoints(); j++) {
		for (int k = 0; k < numberOfObjectives; k++) {
			// ...
		}
		// ...
	}
	// ...
}
```
Same thing for all your `espX` variables. Don't declare stuff before you need it. Otherwise you force the JVM to maintain resources alive while you actually don't need them + you take additional risks of facing unwanted side effects. Let the JVM manages the resources. It is its job and it does it better than you.

# Closing Word: sorry but...

I will finish on that: now I have 63 warnings about raw types. I know I am pushy with generics, but you ask for review, and this is the kind of thing which sums up quickly over time and you already know that (it was 24 yesterday, it's 63 today, and it was several hundreds before I dealt with them). It will continue to increase, and I don't think that you will find more reasons to deal with them later than you have now. So if you postpone, it is probably postponed indefinitely, and I don't plan to fix everyone of them. I don't believe neither that you will magically find some free time at some points to think "oh, nice, let's deal with the warnings now!". Especially when they sum up to 600 after a few months of coding. If you don't do it now, it will just remain here until *I* do it. And by *now*, I mean *NOW*, at the very moment the warning appears, and just after you have finished reading this post.

If I continue to do it, you will also continue to *not* do it. I highly recommend you to setup your IDE to make them appear as errors, so you are sure to *do* it immediately. Maven can still compile, only your IDE will refuse to.

matthieu-vergne:
By the way, I am going now, so I will wait anyway for Monday to see what you thought about all of that.

Have a nice weekend.

ajnebro:
> I did not look for your inability to run your code, which could take some time if it is tricky (I am used to run jars, not class files which need to set the classpath and so on), so I let you deal with that once you have rest and I will focus on the reviewing. 

It is working now. The right command is:
```
mvn -pl jmetal-exec exec:java -Dexec.mainClass="org.uma.jmetal.qualityIndicator.RunIndicator"
```

ajnebro:
> I highly recommend you to setup your IDE to make them appear as errors

I strongly agree. I have just setup Eclipse to do that.

ajnebro:
I like the idea of the `FrontNormalizer` class, but I have to think about this.

The issue is that currently we have:
```java
Hypervolume<List<DoubleSolution>> hypervolume = new Hypervolume(referenceFront);

Double hvValue = hypervolume.evaluate(population);      
Double hvValueNormalized = hypervolume.setNormalize(true).evaluate(population);
```

By using the hypervolume in this way we ensure that both the reference front and front to evaluate are normalized or not. The point is that the front to evaluate must be normalized with respect to the reference front.

This code is not valid  :
> ``` java
>Hypervolume<List<DoubleSolution>> hypervolume = new Hypervolume()
>
>List<DoubleSolution> normalizedPopulation = new FrontNormalizer(referenceFront).normalize(population);
>
>JMetalLogger.logger.info("Hypervolume (N) : " + hypervolume.evaluate(normalizedPopulation));
>JMetalLogger.logger.info("Hypervolume     : " + hypervolume.evaluate(population));
```

because the hypervolume (and most of indicators) needs the reference front to evaluate the population.

**Comment by matthieu-vergne** [comment]
Are you sure? Looking at the code of the hypervolume, I see that the only use you do of the reference front, out of the normalization, is with this line:
```
int numberOfObjectives = trueParetoFront.getPoint(0).getNumberOfDimensions() ;
```
Can't you get this exactly same data from the evaluated front directly? Isn't the number of dimensions supposed to be the same between the front and the reference front? Or did I miss any other use of the reference front?

**Comment by matthieu-vergne** [comment]
I see that the espilon case is more tricky, because you need to know the number of points of the reference front. So I get your point. Maybe some indicators can be refined so you don't need anything in the constructor, but not all. For them we have to see how it can be done. But in general, if we want to achieve a proper genericity of the code, it is hard to do without ensuring that we can use an empty constructor and directly after use the `evaluate()` method. This is the genericity challenge here.

**Comment by matthieu-vergne** [comment]
If we setup some conventions, then we can arbitrarily setup any method to generically retrieve the indicators by using the right constructor. It makes it less controlled (there is conventions to setup, people have to respect them, etc.), but technically speaking there is no problem.

For instance, we could have an additional method like `QualityIndicatorFactory.getReferenceBasedIncidators(Front referenceFront)` which would do the same but look for classes which have no empty constructor but have a constructor taking a `Front` in argument, although it would not make the difference with an indicator which takes a front for another purpose. Another convention would be to still have an empty constructor, but have an explicit method `setReferenceFront(Front referenceFront)`, which is more reliable. The strictest would be to implement a specialized interface which provides such a method, but it does not provide more feature and yet reduce the flexibility (with the second alternative, you can still implement such an interface, but it is not mandatory as long as you have the required method), so I do not recommend it. The interface could be provided to simplify the implementation, but the restriction for retrieval is not recommended.

**Comment by matthieu-vergne** [comment]
And just to be clear: the advantage of the constructor alternative, despite its inconvenient of not being able to make the difference with other types of uses, is that it allows to use "static" indicators (as opposed to "dynamic", not the Java keyword), where you instantiate the indicator with some values which can be setup as `final` members, so you can optimize the computation because you know that it will remain the same.

**Comment by ajnebro** [comment]
> Are you sure? Looking at the code of the hypervolume, I see that the only use you do of the reference front, out of the normalization, is with this line:
```
int numberOfObjectives = trueParetoFront.getPoint(0).getNumberOfDimensions() ;
```
Can't you get this exactly same data from the evaluated front directly (or another way) ? Isn't the number of dimensions supposed to be the same between the front and the reference front? Or did I miss any other use of the reference front?

Sure, do not take only the hypervolume as a reference of quality indicators. In fact, the there is a lot to do with it. I don't see if you have noticed that we have this package:
https://github.com/jMetal/jMetal/tree/master/jmetal-core/src/main/java/org/uma/jmetal/qualityindicator/util

The hypervolume is a complex issue (there are some PhD thesis about it), because computing it is costly when you use it repeadly in an algorithm. There is the "oficial" implementation of E. Zitzler, which is the one in jMetal, and we tried to include another implementation (http://www.wfg.csse.uwa.edu.au/hypervolume/), but it is not working now. 

To compute the hypervolume only a reference point is needed, and the general approach is to calculate if from a reference front. I missing feature is to allow to specify the reference point explicitly in the `Hypervolume` class.

**Comment by ajnebro** [comment]
> I see that the espilon case is more tricky, because you need to know the (number of) points of the reference front. So I get your point. Maybe some indicators can be refined so you don't need anything in the constructor, but not all. For them we have to see how it can be done. But in general, if we want to achieve a proper genericity of the code, it is hard to do without ensuring that we can use an empty constructor and directly after use the `evaluate()` method. This is the genericity challenge here.

I don't know if you remember that the former approach to compute the quality indicators was to have an `evaluate()` method having as parameters the current front and the reference front. This would the be most familiar approach to most of jMetal users, but I liked your idea of including the reference front in the constructor to avoid errors when altering the order of the two fronts when calling `evaluate()`.

**Comment by matthieu-vergne** [comment]
If there is a significant amount of measures which need specific kinds of data, then it could be relevant to forget about a generic `getAvailableIndicators()` or alike. But rather to think about groups of indicators, in which you have an explicit list of indicators, and to which you provide the relevant data for initial configuration.

I think that for now we should stay on a manual instantiation and that grouped retrieval should be added to the todolist for after the 5.0 release. It is not trivial and it needs a proper analysis of the current indicators to identify the fundamental properties to consider for generic coding.

**Comment by ajnebro** [comment]
Yes. I would leave the implementation as is now and later we will have time to rethink about it.

**Comment by matthieu-vergne** [comment]
> I don't know if you remember that the former approach to compute the quality indicators was to have an evaluate() method having as parameters the current front and the reference front. This would the be most familiar approach to most of jMetal users, but I liked your idea of including the reference front in the constructor to avoid errors when altering the order of the two fronts when calling evaluate().

You can solve the order problem by having a setter for the reference front, or to remain with the old idea of providing both of them to the method, by having a specific structure which takes a front to evaluate and a reference front, and you give the structure to `evaluate()` (you basically extract the composition of the two from the indicator). The issue then is really about potential confusions on the use of the indicator + useless verbosity.

If you liked my idea of having `evaluate()` only with the evaluated front just because the reference front can be setup another way to avoid mixing orders, then it is a bit a pity, because I am looking for ontological robustness here. {'^_^}

**Comment by ajnebro** [comment]
> I highly recommend you to setup your IDE to make them appear as errors, so you are sure to do it immediately.

I have set Eclipse and IntellJ Idea to indicate as errors the warnings about raw types, and there is one I cannot solve :(. It happens in lines 42 and 43 of this class: 

https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/operator/impl/crossover/NullCrossover.java

**Comment by matthieu-vergne** [comment]
short answer: don't solve it, add `@SuppressWarnings("unchecked")`. The case of `copy()` is tricky and you should always need to cast wherever you use `copy()`.

Long answer:

This one is because of `copy()`, which anyway is defined to return a `Solution<T>`, not the exact type `S`. For instance, in the case of `DoubleSolution extends Solution<Double>`, any implementation of it will have to implement `Solution<Double> copy()` and not `DoubleSolution copy()`. This is why you still have to cast it here.

You can manually change the returned type of `copy()` when you implement it (so you have the second version instead of the first) because Java allows you to do so (the method signature does not consider the returned type), but it will only affect the specific implementation, not the generic interface (which is the one considered here) so it won't fix it.

If you want to ensure that it returns the right type, then you need to provide to the interface the information about the specific type implemented:
```
public interface Solution<T, S extends Solution<T, ?>> extends Serializable {
  // ...
  public S copy() ;
}
```
The point is that it is not completely controlled (there is still `?` which let it open, and you cannot precise it otherwise you have to repeat indefinitely the same thing) and it makes the use rather cumbersome because you *must* postpone its filling until the very end. For instance, `DoubleSolution extends Solution<Double, DoubleSolution>` would not do because then `copy()` would return a `DoubleSolution`, but when you implement it in `DefaultDoubleSolution` you want it to return a `DefaultDoubleSolution` and not a `DoubleSolution`. To finish you, even if you postpone it to the very last moment, if your class is not `final`, then someone can extend it, and here is the point: he should have a `copy()` which returns its custom type, not your `DefaultDoubleSolution`.

So here you see that it cannot be solved easily, because it relates to structural information that you get at the very end of the hierarchy, which you don't know in advance. If you want to fix that, you will need to think about solution factories/builders or any other class which is separated from the solution itself, so it allows to provide its type as a generic without having this recursivity problem.

ajnebro:
I have to say that genetics are still driving me crazy ...

matthieu-vergne:
{^o^}

Well, it relates to the ability of making generic algorithms, which is not a trivial thing (you have to think about abstract properties rather than concrete cases, which are only examples). It requires to be quite mathematical and rigorous.

Now the point is that I tell you all what I know the way I think it makes sense, but it is my own perspective. You probably did it already, but you would probably benefit from reading other perspectives in order to find one which really speaks to you (or make your own). There is plenty of tutorials about Java generics on the Web, so don't hesitate to look a bit to them. One worth mentioning is surely the [official tutorial from Oracle](https://docs.oracle.com/javase/tutorial/java/generics/), which makes nice tutorials, well split with each section being quite short. Going back to the basis could also help you to clarify some points. I often go back to these tutorials for other stuff that I don't use so often.

ajnebro:
We have to close this issue because we are trapped in more time than expected. 

I have taken the following decisions:
* No need for class `NormalizableQualityIndicators`
* It is responsability of the user to normalize or not the fronts by using class `FrontNormalizer` before computing the indicators (this must be clearly stated in the documentation).
* The constructor of each indicator class has the reference front as an argument.

If we see later a better approach then we'll have a new issue to add to the todo list of jMetal5.1

matthieu-vergne:
Seems a fine trade-off to me.

>The constructor of each indicator class has the reference front as an argument.

*All* indicators? Is it a `QualityIndicator` convention?

ajnebro:
Currently all but `SetCoverage`.

PR #75, opened 2015-06-11
Precise some generics.

matthieu-vergne:


matthieu-vergne:
I close it because it has been redone in #76.

PR #76, opened 2015-06-12
Expliciting generics. Some fixes are inferred.

matthieu-vergne:
With this branch, I focus on expliciting the generics in order to increase type alignments. In the process, expliciting generics create errors which are fixed by expliciting other generics, so the code is only changed regarding these generics, which means it is fundamentally equivalent. This is due to the fact that generics are only used at compilation time: they are lost in the process (type erasure) so the final bytecode is exactly the same with the generics or without, because generics does not apply anymore in the bytecode.

I have actually changed some pieces of the code when needed, but in such a case I notice it in the commit (please check each commit message to know which ones, there is only a few of them). I do not notice them if they are trivial changes like this one:
```
public int compare(Object o1, Object o2) {
  Pair<Integer, Double> pair1 = (Pair<Integer, Double>)o1;
  Pair<Integer, Double> pair2 = (Pair<Integer, Double>)o2;
  ...
```
which has become, due to expliciting types:
```
public int compare(Pair<Integer, Double> pair1, Pair<Integer, Double> pair2) {
  ...
```


matthieu-vergne:
Done. With all these commits, all the generics have been explicited. With Eclipse, I have 258 warnings remaining, compared to 1000+ before.

PR #82, opened 2015-06-13
Ignore mock unchecked generics.

matthieu-vergne:
DON'T MERGE UNLESS YOU KNOW WHAT YOU ARE DOING!

It is not that it can break something, but the assumption here is that we **cannot** fix these warnings, because the mocking is not able to do such a deep check as complying with the generics. If it can, then it is better to do it properly rather than just ignoring these warnings. If not yet done, I suggest you to look how to make `mock()` complies with the parametered types/generics (I never used any mocking solution, so I cannot say anything). If you confirm that you cannot do better, then you can merge this commit.


matthieu-vergne:
Fixed in #139.

Issue #77, opened 2015-06-13
Serializable: why?


matthieu-vergne:
Could you clarify why you made `Problem`, `Algorithm`, `Operator` and many others extend `Serializable`?

While I could understand that you want some data to be stored and retrieved, you basically do it by writing/reading files manually, as far as I understood. Moreover, making problems and algorithms themselves serializable seems to come quite out of the blue. Even more when I see you even add `Serializable` to exception extensions, which are already `Serializable`. It seems to be a kind of automatic behaviour.

matthieu-vergne:
I ask that especially because it generates many warnings, and going ahead thinking about whether to supress them or actually provide a SerialVersionUID, I was just wondering whether or not the correct answer was not actually to remove the extension.

ajnebro:
We are working on a distributed version of jMetal 5.0, and Serialized is required in many classes.

matthieu-vergne:
Can you be more explicit? I saw stuff for multi-threading (so parallelization, not distribution) and serialization is not required there. Do you have something implemented for actual distribution over several machines?

Moreover, serialization is the "stupid" one (not optimized at all, transfer everything), thus is good for trivial/simple cases, but if you want to make a distribution of many things, and stuff as complex as algorithms and cie, you can get far more efficient distribution with externalisable, although it requires manual writing/reading.

matthieu-vergne:
Why closed?

ajnebro:
We need to use Serialized in the development of distributed algorithms. It's sure that `Problem` and `Solution` must be serializable, `Algorithm` probably not. I have to check that.

matthieu-vergne:
From my point of view it is still unclear. It is not really important, because it does not have much impact, but it relates to numerous warnings, and it could be a hint to a bad design. So I remain quite dubious on that point. At least, it is not a priority to me. Maybe I will come back on that when the framework will be well featured (including experiments and GUI).

ajnebro:
Once we close the version 5.0 we have to prepare a todolist for the next release, which should include at least those features of jMetal 4.5 not included in 5.0:
* Configuring algorithms with configuration files (to avoid needing to recompile every time). In jMetal 4.5 we used Java properties; perhaps we could thing in something more expressive, e.g. JSON.
* Support for experimental studies. Now I'm running the algorithms by using a Python script and the results are processed by using jMetal 4.5
* Some algorithm variants are not included, notably FastSMSEMOA and SMPSOhv.

PR #84, opened 2015-06-13
Serialization

matthieu-vergne:
This refers to issue #77: if it is not required to use serializable, then you can merge this pull request. If the serialization is useful, then only the second commit could be taken, although it probably need to be extended to all the other classes of the first commit.


matthieu-vergne:
I close it because I have an updated pull request for that, which is more adapted to the general policy of jMetal.

PR #107, opened 2015-09-04
Possible Null Pointer exception

HaritzPuerto:
Fixed null pointer exception

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/geneticalgorithm/GenerationalGeneticAlgorithm.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/geneticalgorithm/GenerationalGeneticAlgorithm.java
@@ -88,11 +88,10 @@ public class GenerationalGeneticAlgorithm<S extends Solution<?>> extends Abstrac
       parents.add(matingPopulation.get(i + 1));
 
       List<S> offspring = crossoverOperator.execute(parents);
-      mutationOperator.execute(offspring.get(0));
-      mutationOperator.execute(offspring.get(1));
-
-      offspringPopulation.add(offspring.get(0));
-      offspringPopulation.add(offspring.get(1));
+      for(S s: offspring){
+    	  mutationOperator.execute(s);
+    	  offspringPopulation.add(s);
+      }
     }
     return offspringPopulation;
   }


matthieu-vergne:
Why did you close/deleted it?

HaritzPuerto:
I wanted to talk first with Antonio Nebro. I'm one of his students. I'm a newbie in this xD Ok, I'll explain it here. I developed a problem which uses a crossover that I also developed. This only returns one offspring, so it gave me a null pointer exception, so I change it in order to fix it, but I don't know why that class takes the assumption that the crossover must have two offsprings

matthieu-vergne:
Well, the PR is the occasion to start the discussion. Because I think that it makes sense to discard this "2 children" assumption for this generic class. Also it relates to several other discussions where we speak about using iterators instead of indexes (`Collection` instead of `List`). So I would be glad to have someone else to put the finger on this kind of thing, for once. So I could say "You see Antonio, I am not the only one!" instead of a long explanation. {^o^}

matthieu-vergne:
Wow, that was a short discussion. {^o^}

ajnebro:
I agree in that all the for loops should be revised;), not to mention that perhaps we should consider whether adopting the full Java 8 features (I'm thinking in streams) better sooner than later.

matthieu-vergne:
Well, as you may know, Java 8 is not yet broadly available (e.g. the last LTS Ubuntu, [14.04](https://wiki.ubuntu.com/LTS), does not provide OpenJDK although [it is asked since a while to be backported](https://bugs.launchpad.net/trusty-backports/+bug/1368094)). I think that the next LTS is 16.04 (so April 2016), and although it should provide it, it may be still too soon to switch to Java 8 now.

But there is already things that we can do with Java 7 (e.g. it allows to manage efficiently big directories of files) and in the absolute streams are not "absent" from previous versions: we can implement our own stuff already, for what we need, then switch to more standard implementations when we use Java 8.

If you have specific pieces of code that you think to revise (with streams or other stuff), you may open dedicated issues on them, so we could exchange ideas on how to do it with Java 7.

inaciomdrs:
Pertinent changing. Many bio-inspired optimization algorithms have surged these times, and generating two offspring solutions is almost "no longer an standard" for these kind of algorithms, so this kind of flexibility is very necessary not also for mutation operator, but for crossover, selection and any other new kinds of operators used in some optimization methods.

PR #108, opened 2015-09-09
Using constraint with MultithreadedSolutionListEvaluator

jacksonpradolima:
The MultithreadedEvaluator wasn't calling the "evaluateConstraints" method from problem.

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/util/parallel/impl/MultithreadedEvaluator.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/util/parallel/impl/MultithreadedEvaluator.java
@@ -17,6 +17,7 @@ import org.uma.jmetal.problem.Problem;
 import org.uma.jmetal.solution.Solution;
 import org.uma.jmetal.util.JMetalLogger;
 import org.uma.jmetal.util.parallel.SynchronousParallelTaskExecutor;
+import org.uma.jmetal.problem.ConstrainedProblem;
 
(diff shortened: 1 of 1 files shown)


matthieu-vergne:
It seems that the first commit has nothing to do with it. It would be nice to clean this branch, otherwise Antonio pay attention to what it brings before to merge it.

Regarding the second commit, which seems to be the heart of the problem, I have to say that practically speaking it makes sense, but at the same time it is a hint that the architecture is flawed here. Because we need to know about the specificities of a subtype in order to be compatible with it, so the code is not generic enough. So this commit may solve the problem of *`ConstrainedProblem` not being properly used by `MultithreadedEvaluator`*, but it surely does not solve the problem for any potential use of `ConstrainedProblem`, which is the one that we should target to avoid a similar problem to occur again.

So rather than a solution, it appears to me as a trick, and the analysis did not go deep enough to actually solve the problem. And because of that, and because of the additional dependency it requires, I would not integrate that commit and think deeper about what should be done to solve the problem without adding dependencies.

jacksonpradolima:
Hello Matthieu,
Sorry, really one commit is nothing about the really problem.
I done a fork for adjust this problem, but I guess that I already done a fork before.

About the adjust I implemented the same code from SequentialSolutionListEvaluator for maintain the pattern. But, I agree about create something using a correct pattern about it.

ajnebro:
I have just opened an issue to discuss about evaluating constraints: #109

PR #117, opened 2015-10-13
Introduce initialization methods

eug:
This update introduces initialization methods for variables and objective values in some default solutions implementations. Also, it reuses the inherited method ```initializeObjectiveValues()```. Overall this update will make these implementations more consistent.

I have only one concern about the original implementation of ```initializeDoubleVariables()``` in ```DefaultDoubleSolution``` which is sligly different from ```DefaultDoubleBinarySolution```. In this update I have forced both implementations to be identical for consistency reasons. I'm not sure if there is a case where it can produce divergent results. Let me know if I missunderstood something.

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/solution/impl/DefaultDoubleSolution.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/solution/impl/DefaultDoubleSolution.java
@@ -23,22 +23,19 @@ import java.util.HashMap;
  *
  * @author Antonio J. Nebro <antonio@lcc.uma.es>
  */
-public class DefaultDoubleSolution extends AbstractGenericSolution<Double, DoubleProblem>
+public class DefaultDoubleSolution 
+    extends AbstractGenericSolution<Double, DoubleProblem>
     implements DoubleSolution {
+
   /** Constructor */
   public DefaultDoubleSolution(DoubleProblem problem) {
     super(problem) ;
+    
     overallConstraintViolationDegree = 0.0 ;
     numberOfViolatedConstraints = 0 ;
-
-    for (int i = 0 ; i < problem.getNumberOfVariables(); i++) {
-      Double value = randomGenerator.nextDouble() * (getUpperBound(i) - getLowerBound(i)) + getLowerBound(i);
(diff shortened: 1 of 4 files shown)


jmetal-core/src/main/java/org/uma/jmetal/solution/impl/DefaultDoubleSolution.java
```diff
@@ -77,4 +74,11 @@ public DefaultDoubleSolution copy() {
   public String getVariableValueString(int index) {
     return getVariableValue(index).toString() ;
   }
+  
+  private void initializeDoubleVariables() {
+    for (int i = 0 ; i < problem.getNumberOfVariables(); i++) {
+      Double value = randomGenerator.nextDouble(getLowerBound(i), getUpperBound(i)) ;
```
matthieu-vergne: Here you remove the lower bound, which is wrong unless this lower bound is zero. If there is a lower and upper bound, then the value should fall in between, which is achieved by computing "min + rand x (max - min)", so if rand = 0 the result is min, if it is 1 the result is max, and otherwise it is in between.

matthieu-vergne: No, it's OK, I did not notice the actual difference is the function, not the formula.

nextDouble() without arguments returns a value in [0;1], while nextDouble(min, max) returns a value between min and max, so the result is the same at the end. Well actually I did not check but this is the usual semantic, so I guess it is the same here.

matthieu-vergne:
If the implementation builds on a usual semantic, it should be equivalent. So from my point of view, your code is better and if there is a mistake it probably comes from the implementation of the random generator.

jmetal-core/src/main/java/org/uma/jmetal/solution/impl/DefaultDoubleSolution.java
eug: Yes, that's was what I thought. Also, the ```DefaultIntegerDoubleSolution``` is implemented in the same fashion. So, I don't think this will be a problem.

eug:
Agreed.

PR #136, opened 2016-03-11
Fix for odd population size

samokhine:
To fix "array out of index" exception in case of odd population size.


samokhine:
Well... I believe that my original pull requests is good then and can be just merged. Please let me know if you disagree and I'll tweak it.

PR #137, opened 2016-03-24
Issue with generics

matthieu-vergne:
The first commit is just a useless cast.

The second tries to fix some $#!?§£... hum, 36 raw type warnings, but I am facing a conflict.

At [line 105 of BinaryProblemStudy](https://github.com/jMetal/jMetal/blob/master/jmetal-exec/src/main/java/org/uma/jmetal/experiment/BinaryProblemsStudy.java#L105) you compute the quality indicators of your experiment. This experiment, defined at [line 85](https://github.com/jMetal/jMetal/blob/master/jmetal-exec/src/main/java/org/uma/jmetal/experiment/BinaryProblemsStudy.java#L85), deals with `BinarySolution` solutions and a `List<BinarySolution>` for its result. To compute them, you instantiate a `ComputeQualityIndicators` by providing it this experiment. Now, this is the tricky part: [it does not fill its generics](https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/ComputeQualityIndicators.java#L56) (Boo! Bad boy!) but what you get is just a warning, so let's just add a cool `@SuppressWarnings("unchecked")` at [line 51](https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/ComputeQualityIndicators.java#L51), so no other warning appears (No warning? No problem!). Then let's continue: in its `run()` method, you have many things which do not, also, fill their generics (Boo! But they are not displayed anymore, so who cares.). In particular at [line 63](https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/ComputeQualityIndicators.java#L63) where you obtain the indicator from the experiment, you should then obtain a `GenericIndicator<BinarySolution>`, but here the code does not know because no generics (Shhh! Don't care!). If we go ahead, we reach [line 96](https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/ComputeQualityIndicators.java#L96) which generates a front as a `List<DoubleSolution>` and then use it on the next line to be evaluated by the indicator... Oops! Where is our `BinarySolution`?

Consequently, I am currently fixing these warnings (which are errors for me, so no `unchecked` can hide them: no excuse for lazy guys!) and I face this situation where it cannot compile, because the types are conflicting. So I kindly ask someone to tell me how to fix this issue or to give it a thorough check (and to switch the raw type warnings to errors for people not having it yet).

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/util/distance/impl/EuclideanDistanceBetweenSolutionAndASolutionListInObjectiveSpace.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/util/distance/impl/EuclideanDistanceBetweenSolutionAndASolutionListInObjectiveSpace.java
@@ -27,10 +27,10 @@ import java.util.List;
 public class EuclideanDistanceBetweenSolutionAndASolutionListInObjectiveSpace<S extends Solution<Double>, L extends List<S>>
     implements Distance<S, L> {
 
-  private EuclideanDistanceBetweenSolutionsInObjectiveSpace distance ;
+  private EuclideanDistanceBetweenSolutionsInObjectiveSpace<S> distance ;
 
   public EuclideanDistanceBetweenSolutionAndASolutionListInObjectiveSpace() {
-    distance = new EuclideanDistanceBetweenSolutionsInObjectiveSpace() ;
+    distance = new EuclideanDistanceBetweenSolutionsInObjectiveSpace<S>() ;
   }
 
   @Override
--- a/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/ComputeQualityIndicators.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/ComputeQualityIndicators.java
@@ -48,22 +48,21 @@ import java.util.*;
  *
  * @author Antonio J. Nebro <antonio@lcc.uma.es>
  */
-@SuppressWarnings("unchecked")
 public class ComputeQualityIndicators<Result> implements ExperimentComponent {
 
   private final Experiment<?, Result> experiment;
 
-  public ComputeQualityIndicators(Experiment experiment) {
+  public ComputeQualityIndicators(Experiment<?, Result> experiment) {
     this.experiment = experiment ;
     this.experiment.removeDuplicatedAlgorithms();
   }
 
   @Override
   public void run() throws IOException {
-    for (GenericIndicator indicator : experiment.getIndicatorList()) {
+    for (GenericIndicator<?> indicator : experiment.getIndicatorList()) {
       JMetalLogger.logger.info("Computing indicator: " + indicator.getName()); ;
 
-      for (TaggedAlgorithm algorithm : experiment.getAlgorithmList()) {
+      for (TaggedAlgorithm<Result> algorithm : experiment.getAlgorithmList()) {
         String algorithmDirectory ;
         algorithmDirectory = experiment.getExperimentBaseDirectory() + "/data/" +
             algorithm.getTag() ;
@@ -146,8 +145,8 @@ public class ComputeQualityIndicators<Result> implements ExperimentComponent {
   }
 
   public void findBestIndicatorFronts(Experiment<?, Result> experiment) throws IOException {
-    for (GenericIndicator indicator : experiment.getIndicatorList()) {
-      for (TaggedAlgorithm algorithm : experiment.getAlgorithmList()) {
+    for (GenericIndicator<?> indicator : experiment.getIndicatorList()) {
+      for (TaggedAlgorithm<Result> algorithm : experiment.getAlgorithmList()) {
         String algorithmDirectory;
         algorithmDirectory = experiment.getExperimentBaseDirectory() + "/data/" +
             algorithm.getTag();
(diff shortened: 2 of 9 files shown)


ajnebro:
I'm happy to see that you are enyoing this :P.

Jokes apart, I also dislike having warnings, and I don't like to use tricks to hide them. Just consider that I needed the experiment features to be used in several papers with firm deadlines and my goal was to have a fast working implementation, not to spend time in designing the best approach to the experiments. In fact, when doing this I have found some casting issues that must be addressed. 

Now I'm going to take a deep look to your pull request.

matthieu-vergne:
Don't care about the form. I am French, so I complain. {^_^}
Just take the meat.

ajnebro:
>  If we go ahead, we reach line 96 which generates a front as a List<DoubleSolution> and then use it on the next line to be evaluated by the indicator... Oops! Where is our BinarySolution?

Ok, let's go with this issue. The point is that quality indicators are applied to Pareto front approximations, i.e., the generated "FUNxx" files, which are composed of rows of double values. These files are produced when running the algorithms  (`new ExecuteAlgorithms<>(experiment).run();`), which is the first step in any experiment. 

After that, to compute the quality indicators the FUNxx files must be read and converted into a list of solutions. However, there is no way to know about the original solution type only from the FUNxx files, so we use this sentence: 
``` java
List<DoubleSolution> normalizedPopulation = FrontUtils.convertFrontToSolutionList(normalizedFront) ;
```
which creates a dummy list of DoubleSolution. This list is only used to get the objective values of its solutions.

matthieu-vergne:
OK, so if I have well understood, you retrieve always a list of `DoubleSolution` independently of the type used initially. I guess it is because this is the most generic type (technically speaking, any solution can be represented as a `DoubleSolution`). Conceptually speaking, there is obviously a problem here, which is that solutions should have an homogeneous type.

One solution is to store the type of solution into your files, because after all your files are here to store the information about your solutions so you can retrieve them. We could exploit the serialization, but we may face some difficulties because generics are normally forgotten at runtime, so we could need to implement some custom process to retrieve the type anyway. Consequently I think it would be better to revise your file format to just integrate this information (it seems to me to be a custom format, so you can adapt it the way you want, right?). Typically, you "just" store the full class name as first line, and when you read it you "just" instantiate the solution using reflexion methods. what makes it not "just" like that is the instanciation process of the solution: if you need constructor arguments or if you need to use setters, then you need to store all this information for you to be able to retrieve it. If all your solution classes are standardized, then the class name should be sufficient.

A second solution, which removes a bit of burden, is to manage it through the code by providing explicitly the class to use for instanciating solutions. The process is basically the same than above, but you don't have to think about how to write/read it in your file. I think you should be able to use this solution because you run an experiment or read its results by knowing which solution type has been used (at least the problem, which anyway provides the solution type), so you should be able to provide such information through the code I guess. But again the complexity depends on the complexity of the instanciation.

The last solution I see is to actually have an architecture which makes any solution a child of `DoubleSolution`, so you can just use it as the general type anywhere. I won't give details on that, because I strongly disagree with such an approach: `Solution` is supposed to be the most genreic type, and if a spceific type is not taken as the most generic one it is precisely because a solution can encode anything. For instance, I never use solutions which use simple numbers, but structures, so this kind of architecture falls short as soon as you face some complex situation for which simple numbers are not suited.

The easiest one seem to me the second solution, while the first solution would be preferred but I am not sure it is worth to put that much effort into it, considering that we are revising the experiments. But it would be a good exercise, so I may also recommend it for that purpose (we will probably need something like that anyway).

ajnebro:
> OK, so if I have well understood, you retrieve always a list of DoubleSolution independently of the type used initially. 

Yes.

> I guess it is because this is the most generic type (technically speaking, any solution can be represented as a DoubleSolution).

No. As only the objective values are needed to compute quality indicators any kind of solution would be valid. The choice of DoubleSolution is somehow arbitrary. 

Think that the FUNxx files only contain objective values. This way their contents can be plotted directly with Gnuplot, R or whatever tool.

ajnebro:
We have to fix this soon because I accepted your pull request before reading your comments and now the master branch do not compile :(

matthieu-vergne:
You got what you deserved. Next time you will think to check a pull request before to merge it naively. {^_^}

> As only the objective values are needed to compute quality indicators any kind of solution would be valid. The choice of DoubleSolution is somehow arbitrary.

You mean that these are only the values to maximize/minimize, which are always `double` independently of what the genes are, while it is the latter which constrain the solution type. So any solution type is fine if we only need to retrieve these values. Am I right?

If yes, then there is a clear dichotomy to do here: what you need to recover is **not** a solution, but something less than that. So the best alternative seems to me to properly design your retrieval of objective values as a generic `List<Double>` or something like that.

matthieu-vergne:
Actually more a `List<List<Double>>` because there is several values for each solution, or if you want to name the values based on what they represnt `List<Map<String, Double>>`. Something like that.

ajnebro:
> You got what you deserved. Next time you will think to check a pull request before to merge it naively. {^_^}

I know ;)

ajnebro:
> Actually more a List<List<Double>> because there is several values for each solution, or if you want to name the values based on what they represent List<Map<String, Double>>. Something like that.

Hmmm .... yes. The idea of using a kind of dummy DoubleSolution was because the quality indicators are applied to lists of Solution. In fact, we defined a `Front` class for similar a purpose.

matthieu-vergne:
Maybe quality indicators, if they never relate to the genes but only to the objective values, should be redesigned correspondingly (personally I don't think so, but it can be argued).

ajnebro:
I'm going to think about it. 

The compiling error can be converted into an unchecked warning. Let it be for now while I see the issue of the quality indicators.

I have to take a look also to your experiment architecture.

matthieu-vergne:
Let the experiment architecture for later, or just for inspiring yourself. I don't know yet if it is the one I will suggest, because other things need to be changed first.

PR #139, opened 2016-03-26
Fix warnings

matthieu-vergne:


matthieu-vergne:
For the second commit, it is similar to #84 but I only suppress the warnings, I do not remove any `Serializable`. The point here is that the code does exactly the same thing, we just explicitly say "Yes I know, but I don't care", which already what is done. The reason is that, so far, it seems you did not need to provide dedicated `serialVersionUID`, otherwise you already did it in the classes requiring it. So for all the remaining classes, I assume that it is confirmed you do not need it.

I often do this in my code, so it is not just a lazy choice: if you always use serialization in a context which is prone to provide the same `serialVersionUID` then you don't need to provide it yourself, and doing so could hide some problems. So as long as you don't face `ClassCastException` due to wrong serialization versions, you can just assume that it is OK, and think more about it when you face this exception.

matthieu-vergne:
For the third commit, there is some changes in the code:
- Use `ArrayList` instead of simple array, which should be equivalent in terms of performance.
- Add few generics to be able to precise which object to take, which allows to replace some casts by a proper enforcement.

All the rest is just removal of warnings when this is an external function (e.g. `mock()`) which does not allow to dig in, or when we deal with an heterogeneous array. For this latter, it could be replaced by a proper structure, but I did not go so far.

matthieu-vergne:
In total, only 60 warnings remain:
- 2 come from my own classes that I would like to keep (`SolutionBuilder` and `SolutionEvaluator` are experimental stuff yet to be considered).
- the other 58 are unnused stuff that you have to go through to see if it should be actually used or if you can remove them. Probably a lot should be removed if not all, so it should be quick to fix.

PR #140, opened 2016-04-02
Minor improvements.

matthieu-vergne:
Again some warning fixed. I also re-did what I did before for `DominanceRanking`, but properly ran the tests to see the bug. My bad not doing it before. So it should be OK now.


ajnebro:
Certainly the `DominanceRanking` class was not working properly since the last pull request acceptance. I restored the previous version to make it work again.

I observed then that the `DominanceRankingTest` contained only a test case. I have extended it to get a 100% coverage.

Now, GitHub says that this pull request has conflicts to be resolved. I'll take a look to see what happens.

matthieu-vergne:
Well, probably because you modified some stuff close to my own modifications in your last commit (e.g. Yebisu removed a commented line in `DominanceRanking`). Should not be hard to align I think. I can look at it and re-align if you want.

PR #146, opened 2016-05-31
Adjust in the method crowdingDistanceSelection for NSGAII and NSGAII45

jacksonpradolima:
Hello,
I came across a bug where the crowding distance was not calculated for all solutions only for the last subfront. In the next iteration, the selection operator would some times use an empty crowding distance value in the binary tournament, which would cause a random selection.

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/spea2/SPEA2Builder.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/spea2/SPEA2Builder.java
@@ -28,14 +28,16 @@ import java.util.List;
 
 /**
  * @author Juan J. Durillo
+ * @param <S>
  */
 public class SPEA2Builder<S extends Solution<?>> implements AlgorithmBuilder<SPEA2<S>> {
   /**
    * SPEA2Builder class
    */
   protected final Problem<S> problem;
-  protected int maxIterations;
+  protected int maxEvaluations;
   protected int populationSize;
+  protected int archiveSize;
   protected CrossoverOperator<S> crossoverOperator;
   protected MutationOperator<S> mutationOperator;
   protected SelectionOperator<List<S>, S> selectionOperator;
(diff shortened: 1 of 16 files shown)


matthieu-vergne:
Considering that it is a fix, it would be nice to have a test to highlight it (which passes with your changes but not without) so we can avoid regressions.

ajnebro:
Hi Jackson.
  Thanks for the pull request. It seems that you are right by I'm going to revise code before accepting it.
  
  Have you noticed some improvement in some algorithm after making the modification?


Antonio

fritsche:
Hi,

I have run some tests, and it looks like the problem doesn't affect much the algorithm performance. But it would be necessary run some statistical test to be sure. 
For DTLZ1-7 and WFG1-4 with two objectives, I have computed the count of how many times one version have best hypervolume values than the other. (Being NSGA-IIa the version with the modification).

````
Best hypervolume (count of 30 runs)
Problem	NSGA-IIa	NSGA-II
DTLZ1	16	        14
DTLZ2	18	        12
DTLZ3	20	        10
DTLZ4	11	        19
DTLZ5	16	        14
DTLZ6	15	        15
DTLZ7	14	        16
WFG1	20	        10
WFG2	16	        14
WFG3	14	        16
WFG4	16	        14
````
Besides, **there is still a similar situation on the first iteration**. None of the solutions have the CrowdingDistance or the Dominance Depth values because it is computed only on the replacement step. So, **at the first iteration, the selection is completely random**.

**Comment by ajnebro** [comment]
> Besides, there is still a similar situation on the first iteration. None of the solutions have the CrowdingDistance or the Dominance Depth values because it is computed only on the replacement step. So, at the first iteration, the selection is completely random.

Yes, I was aware of that :)

**Comment by matthieu-vergne** [comment]
The question being: should it be fixed too or is it done on purpose?

**Comment by ajnebro** [comment]
I'm taking a look to the code. The point is that NSGA-II, for example, uses a `RankingAndCrowdingSelection` operator, which computes the ranking and crowding before selecting a number of solutions, so never will be a comparison of solutions having no crowding distance assignment.

However, if the replacement in NSGA-II computes also the ranking and crowding, it seems at a first glance that the selection operator is duplicating a computation, given that replacement is the last step of a generation and selection the first step in the next generation.

Still reviewing the code ...

**Comment by matthieu-vergne** [comment]
Does that mean that the single responsibility principle is not satisfied?

**Comment by ajnebro** [comment]
It's sure that the principle is not satisfied by the `RankingAndCrowdingSelection`, which computes the ranking and crowding and the select a number of solutions. I'm thinking in removing the first step and set the precondition that the operator assumes that the list of solutions has been previously ranked.

**Comment by matthieu-vergne** [comment]
Why commenting the GPG plugin?

**Comment by jacksonpradolima** [comment]
You can disregard this change, I did for my fork. This comment is because appear some errors because of GPG plugin when I do "mvn install". 
I know that do "mvn package" is the way,  however, I make "mvn install" because I wanna install the package into the local repository, for use as a dependency in other projects locally :)
(Sorry about my English)

File: jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/spea2/SPEA2Builder.java [hunk 1]
```diff
@@ -28,38 +28,43 @@
...
  * @author Juan J. Durillo
+ * @param <S>
  */
 public class SPEA2Builder<S extends Solution<?>> implements AlgorithmBuilder<SPEA2<S>> {
   /**
    * SPEA2Builder class
    */
   protected final Problem<S> problem;
-  protected int maxIterations;
+  protected int maxEvaluations;
```
matthieu-vergne: The change here is not made for the constructor.

matthieu-vergne: The change here is not done in the constructor of SPEA2.

matthieu-vergne: Fixed in d6861ce3ea16567f8ee0d79936ae30f7d50f99ff

matthieu-vergne:
What about this issue? Did it come to a mergeable state? Is it abandonned?

PR #174, opened 2016-11-16
Remove unused imports.

matthieu-vergne:


matthieu-vergne:
Replaced by request #175

PR #176, opened 2016-11-17
Decrease loops for faster testing.

matthieu-vergne:
There is [a test](https://github.com/jMetal/jMetal/blob/9c110f9c9dec116ba4f40589815158ea8accc991/jmetal-core/src/test/java/org/uma/jmetal/solution/impl/DoubleSolutionComparisonIT.java#L51) which takes really long to execute. It seems that it is just to compare that 1 case takes less time to execute than another. Can't we reduce the amount if it is just for that? Like to 100 or 1'000 instead of 1'500'000? The other test of the same class also takes several seconds by looping 150'000 times.

I have run them dozens of times with 1'000 and it failed only once. So I guess 10'000 is enough. With this value, it takes less than a second per test. With your values, one takes 6s and the other 106s on my machine, so it's really long. All the tests but these two are really fast (running the 556 tests takes less than 5s), so it is frustrating to wait for these two. Especially when you want to check several commits.

### DIFF
--- a/jmetal-core/src/test/java/org/uma/jmetal/solution/impl/DoubleSolutionComparisonIT.java
+++ b/jmetal-core/src/test/java/org/uma/jmetal/solution/impl/DoubleSolutionComparisonIT.java
@@ -28,13 +28,14 @@ public class DoubleSolutionComparisonIT {
     long startTime ;
 
     startTime = System.currentTimeMillis() ;
-    for (int i = 0 ; i < 150000; i++) {
+    int loops = 10000;
+	for (int i = 0 ; i < loops; i++) {
       problem1.createSolution() ;
     }
     long timeToCreateObjectsWithProblem1 = System.currentTimeMillis() - startTime ;
 
     startTime = System.currentTimeMillis() ;
-    for (int i = 0 ; i < 150000; i++) {
+    for (int i = 0 ; i < loops; i++) {
       problem2.createSolution() ;
     }
     long timeToCreateObjectsWithProblem2 = System.currentTimeMillis() - startTime ;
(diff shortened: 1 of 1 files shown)


matthieu-vergne:
By the way, even with huge values it fails some times. It just failed for me.

ajnebro:
That class include some integration tests, so they shouldn't be executed when running the unit tests. 

They are intented to check that an alternative implementation of `DoubleSolution` called [`ArrayDoubleList`](https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/solution/impl/ArrayDoubleSolution.java), which uses a double[] for storing the variables, is more efficient that the default implementation based on List<Double>.

matthieu-vergne:
And how do you run everything but them without annotating them as ignored?

ajnebro:
I use Maven indicating different goals:
* `mvn test` (unit tests)
* `mvn integration-test `

matthieu-vergne:
Seems I have to look at how to manage it in Eclipse.

ajnebro:
I always use Maven from the command line, but it's sure that you can run Maven in Eclipse.

matthieu-vergne:
But even Maven in Eclipse has its drawbacks:
- it also takes time to recompile while Eclipse already did the job,
- it is shown through textual logs, so you have to go back in the logs to know what went wrong,
- if something fails, the log gives you a link... to the exception stack which has been logged (pure text), which means that you have to read that stack and browse manually to reach the file/method/assertion which went wrong.

At the opposite, the "Run as JUnit test" feature of Eclipse:
- directly runs the tests,
- runs them in a graphical interface which can show only what went wrong,
- allows to click on it to reach directly to the faulty assertion.

It's hard to go back to text and manual browsing when you can figure out something went wrong and go to the problem in 1s and 2 clicks.

matthieu-vergne:
Although the time reduction argument still holds (I hardly see how running them a million time is so useful), the split between unit and integration tests should be effective. So I propose an utilitary suite in the project for people like me who would not pass through Maven (#178).

Issue #166, opened 2016-09-23
Time to move to Java 8


ajnebro:
Until now we have avoided to use Java 8 features in jMetal. The reason, in my case, was that we run jMetal code in Hadoop systems, which only supported Java 7, but now our Hadoop cluster has Java 8. 

So, unless someone give some reason against that, my idea is to start to use lambdas, streams, etc progressively.

matthieu-vergne:
The last Ubuntu (16.04) also provides Java 8 in its repositories (through OpenJDK). So I don't see any reason to avoid it anymore.

ajnebro:
Great :)

ajnebro:
I'm thinking on how to proceed to adapt code to Java 8 (in re-reading again "Java 8 in Action").

Instead of merely changing particular pieces of code when we see the oportunity of doing so, perhaps it would be a better idea to start reviewing the core classes, indicating in this issue the change we intend to do, discussing it, and after aproving it, commit the modification.

matthieu-vergne:
I would suggest you to go a step further: use this issue to discuss and list the different features of Java 8 that could be exploited in jMetal, and open a dedicated issue for each of them. In these dedicated issues, we can:
- discuss how to exploit the feature in jMetal
- agree on how to reliably identify the pieces of code to update
- agree on how to reliably update them
- provide commits for the issue

Once such a specific issue is closed, you can check the corresponding point in the list here. The list can be updated progressively if we find out other intereting features to exploit. Once the full list is checked, the new version can be released.

ajnebro:
First version of the list of features:
* Default methods in interfaces
* Lamba expressions
* Streams
* Optional instead of null

matthieu-vergne:
Based on the [list of features provided by Oracle](http://www.oracle.com/technetwork/java/javase/8-whats-new-2157071.html), what I think might be of interest to investigate:
- [Lambda expressions](https://docs.oracle.com/javase/tutorial/java/javaOO/lambdaexpressions.html), especially because we can use plenty of single-method interfaces in jMetal by just having their definitions (lambdas allow to "implement them" without actually doing so). Listeners are of particular interest.
[...]

PR #177, opened 2016-11-17
Defaults and lambdas

matthieu-vergne:
Commits related to Java 8 adaptation (#166).

After playing around with lambdas and default methods, here are some commits. There is also a bit of analysis to share:
- default methods are nice to transform abstract classes into interfaces, which allows then to implements several of them instead of extending only one. However, if the abstract class needs internal variables, then it must remain abstract (interface can't have that, only `static final` variables). In other words, default methods can be used as long as the methods are relatively independent or stateless. There is probaly ways to hack that, but it seems to be the general way to go for them.
- lambdas can be used for any interface which meets the functional requirements: being an interface requiring to implement a single method. "Requiring to implement" is important, because the interface may define several methods but the ones having default methods do not count. For instance, I have revised the `DescribedEntity` interface to use default methods, so any interface extending it is completely free to implement or not there own methods (they should, but technically they can avoid it). Especially, `PullMeasure` and `QualityIndicator` adding a single method each, they can be used through lambdas. Without the default methods they couldn't, because they would need to implement 3 methods instead of 1. A counter example is `DensityEstimator`, which adds a single method but cannot be used through lambdas because it extends `SolutionAttribute`, which adds several methods with no default.

Other classes already usable as lambdas:
- `Distance`
- `ExperimentComponent`
- `Neighborhood`
- `AlgorithmBuilder`

I did not exploit lambdas for them in my commits because they are all used for defining classes. None have anonymous instances, which are the best candidates for using lambdas.

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/measure/impl/CountingMeasure.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/measure/impl/CountingMeasure.java
@@ -150,13 +150,7 @@ public class CountingMeasure extends SimplePushMeasure<Long> implements
 		if (linkedMeasures.containsKey(measure)) {
 			// already linked
 		} else {
-			MeasureListener<T> listener = new MeasureListener<T>() {
-
-				@Override
-				public void measureGenerated(T value) {
-					increment();
-				}
-			};
+			MeasureListener<T> listener = (value) -> increment();
 			measure.register(listener);
 			linkedMeasures.put(measure, listener);
 		}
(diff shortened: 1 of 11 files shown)


matthieu-vergne:
A small comment on the default methods for `DescribedEntity`: it might be arguable to use them, because they do provide default values but they kind of miss the point of default methods I think. The advantage they give here is mainly regarding other interfaces which may add a single method. These interfaces then can be used as lambdas. The point is that, because these lambdas rely on the default methods for the name and description, which are all but informative, you cannot use that for production servers. It is nice for tests, because you can implement some fake measures through lambdas, which makes the code simpler, but out of tests you should use `SimpleDescribedEntity` or something else, but not the default methods. Said another way, these default methods allow a behaviour which normally is not wanted. And regarding the use for tests, it might be that it is just a limited way to make mocks, while it probably miss the features of proper mocking.

So I think that my current use of default methods for the `DescribedEntity` (and the modifications exploiting that) is a kind of abuse here.

matthieu-vergne:
What do you think about what I discuss in my last post? I see you review the changes, but you don't say anything. {'^_^}

PR #186, opened 2017-02-25
Factor redundant code of R2.

matthieu-vergne:


matthieu-vergne:
Seems like these checks do not know what a code factoring is about.

matthieu-vergne:
Replaced by #191.

PR #192, opened 2017-03-11
R2 indicator file load fixed

vinixnan:
The method R2(String file) is not reading all file lines. It finds just the number of objectives but do not reads the lines properly. I changed just this method and added one import.


matthieu-vergne:
Could you add a test which shows the problem? (red without your fix, green with it)
This way we can avoid regressions.

matthieu-vergne:
And please, don't change the indentation in future commits. The diffs is more confusing otherwise.

PR #198, opened 2017-04-08
Random generalisation

matthieu-vergne:
I just completely revised the commits with many additions. New description in the next post.

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/operator/impl/crossover/BLXAlphaCrossover.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/operator/impl/crossover/BLXAlphaCrossover.java
@@ -19,6 +19,7 @@ import org.uma.jmetal.solution.util.RepairDoubleSolution;
 import org.uma.jmetal.solution.util.RepairDoubleSolutionAtBounds;
 import org.uma.jmetal.util.JMetalException;
 import org.uma.jmetal.util.pseudorandom.JMetalRandom;
+import org.uma.jmetal.util.pseudorandom.RandomGenerator;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -37,7 +38,7 @@ public class BLXAlphaCrossover implements CrossoverOperator<DoubleSolution> {
 
   private RepairDoubleSolution solutionRepair ;
 
-  private JMetalRandom randomGenerator ;
+  private RandomGenerator<Double> randomGenerator ;
 
   /** Constructor */
   public BLXAlphaCrossover(double crossoverProbability) {
@@ -51,6 +52,11 @@ public class BLXAlphaCrossover implements CrossoverOperator<DoubleSolution> {
 
   /** Constructor */
   public BLXAlphaCrossover(double crossoverProbability, double alpha, RepairDoubleSolution solutionRepair) {
+	  this(crossoverProbability, alpha, solutionRepair, () -> JMetalRandom.getInstance().nextDouble());
+  }
+
+  /** Constructor */
+  public BLXAlphaCrossover(double crossoverProbability, double alpha, RepairDoubleSolution solutionRepair, RandomGenerator<Double> randomGenerator) {
     if (crossoverProbability < 0) {
       throw new JMetalException("Crossover probability is negative: " + crossoverProbability) ;
     } else if (alpha < 0) {
@@ -59,7 +65,7 @@ public class BLXAlphaCrossover implements CrossoverOperator<DoubleSolution> {
 
     this.crossoverProbability = crossoverProbability ;
     this.alpha = alpha ;
-    randomGenerator = JMetalRandom.getInstance() ;
+    this.randomGenerator = randomGenerator ;
     this.solutionRepair = solutionRepair ;
   }
 
@@ -110,7 +116,7 @@ public class BLXAlphaCrossover implements CrossoverOperator<DoubleSolution> {
     double upperBound;
     double lowerBound;
 
-    if (randomGenerator.nextDouble() <= probability) {
+    if (randomGenerator.getRandomValue() <= probability) {
       for (i = 0; i < parent1.getNumberOfVariables(); i++) {
         upperBound = parent1.getUpperBound(i);
         lowerBound = parent1.getLowerBound(i);
@@ -137,10 +143,10 @@ public class BLXAlphaCrossover implements CrossoverOperator<DoubleSolution> {
         minRange = min - range * alpha;
         maxRange = max + range * alpha;
 
-        random = randomGenerator.nextDouble();
+        random = randomGenerator.getRandomValue();
         valueY1 = minRange + random * (maxRange - minRange);
 
-        random = randomGenerator.nextDouble();
+        random = randomGenerator.getRandomValue();
         valueY2 = minRange + random * (maxRange - minRange);
 
         valueY1 = solutionRepair.repairSolutionVariableValue(valueY1, lowerBound, upperBound) ;
--- a/jmetal-core/src/main/java/org/uma/jmetal/operator/impl/crossover/DifferentialEvolutionCrossover.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/operator/impl/crossover/DifferentialEvolutionCrossover.java
@@ -17,6 +17,7 @@ import org.uma.jmetal.operator.CrossoverOperator;
 import org.uma.jmetal.solution.DoubleSolution;
 import org.uma.jmetal.util.JMetalException;
 import org.uma.jmetal.util.JMetalLogger;
+import org.uma.jmetal.util.pseudorandom.BoundedRandomGenerator;
 import org.uma.jmetal.util.pseudorandom.JMetalRandom;
 
 import java.util.ArrayList;
@@ -56,7 +57,8 @@ public class DifferentialEvolutionCrossover implements CrossoverOperator<DoubleS
 
   private DoubleSolution currentSolution ;
 
-  private JMetalRandom randomGenerator ;
+  private BoundedRandomGenerator<Integer> jRandomGenerator ;
+  private BoundedRandomGenerator<Double> crRandomGenerator ;
 
   /** Constructor */
   public DifferentialEvolutionCrossover() {
@@ -70,12 +72,25 @@ public class DifferentialEvolutionCrossover implements CrossoverOperator<DoubleS
    * @param variant
    */
   public DifferentialEvolutionCrossover(double cr, double f, String variant) {
+	  this(cr, f, variant, (a, b) -> JMetalRandom.getInstance().nextInt(a, b), (a, b) -> JMetalRandom.getInstance().nextDouble(a, b));
+  }
+
+  /**
+   * Constructor
+   * @param cr
+   * @param f
+   * @param variant
+   * @param jRandomGenerator
+   * @param crRandomGenerator
+   */
+  public DifferentialEvolutionCrossover(double cr, double f, String variant, BoundedRandomGenerator<Integer> jRandomGenerator, BoundedRandomGenerator<Double> crRandomGenerator) {
     this.cr = cr;
     this.f = f;
     this.k = DEFAULT_K ;
     this.variant = variant ;
 
-    randomGenerator = JMetalRandom.getInstance() ;
+    this.jRandomGenerator = jRandomGenerator;
+    this.crRandomGenerator = crRandomGenerator ;
   }
 
   /** Constructor */
@@ -128,13 +143,13 @@ public class DifferentialEvolutionCrossover implements CrossoverOperator<DoubleS
     child = (DoubleSolution)currentSolution.copy() ;
 
     int numberOfVariables = parentSolutions.get(0).getNumberOfVariables();
-    jrand = randomGenerator.nextInt(0, numberOfVariables - 1);
+    jrand = jRandomGenerator.getRandomValue(0, numberOfVariables - 1);
 
     // STEP 4. Checking the DE variant
     if (("rand/1/bin".equals(variant)) ||
             "best/1/bin".equals(variant)) {
       for (int j = 0; j < numberOfVariables; j++) {
-        if (randomGenerator.nextDouble(0, 1) < cr || j == jrand) {
+        if (crRandomGenerator.getRandomValue(0.0, 1.0) < cr || j == jrand) {
           double value;
           value = parentSolutions.get(2).getVariableValue(j) + f * (parentSolutions.get(0).getVariableValue(
             j) -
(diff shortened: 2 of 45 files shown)


matthieu-vergne:
Considering that nobody looked it yet, I took the freedom to add many stuff. The two first commits introduce two kinds of generic random generators (functional interfaces). All the others apply them to various classes with different levels of complexity:
- the first one makes trivial changes (`JMetalRandom` to relevant interface).
- the second makes other trivial changes but which have an impact on the tests (especially mocks).
- the third shows how a single `JMetalRandom` instance can be split in several generic generators when different uses are necessary.
- the last one is a slightly different generalisation, which include other aspects.

In the process, I have revised almost all the classes of the core project which use `JMetalRandom`. Only 2 classes have not been revised:
- `AbstractGenericSolution`
- `ArrayDoubleSolution`

I did not revise them because they both have a `protected` random generator, which means that I cannot revise it without taking the risk to break existing implementations elsewhere (including projects using the lib). Notice that this `protected` is useless because of the singleton: any class requiring it can use directly the singleton. In particular, doing so allows to remove the generator of `AbstractGenericSolution` because it is never used there.

matthieu-vergne:
In the previous commits, I have worked systematically to try to ensure that, if a custom generator is provided, it is used in every relevant aspects. The goal was to ensure that all the uses of the singleton can be replaced by a custom generator. To do so, I played with `@Deprecated` to spot the right pieces of code to check. None of these deprecation tags were comitted, only the relevant changes. I followed this process for the modifications:
- make the `JMetalRandom.getInstance()` method deprecated, so all its uses are listed in the IDE as warnings or errors (depending on the IDE configuration).
- change the relevant method/constructor by adding a parameter for the generator,
- adapt the code of the class correspondingly,
- add a method/constructor without the parameter to use the singleton as a default (for legacy code),

Then, I followed this process to see if there was dependencies to check:
- make the additional methods (legacy) deprecated and remove the deprecation from the singleton, so only the impacted methods are listed as warnings or errors,
- check that the listed classes are not part of the modified classes,

For the last point, the assumption is that if a modified class uses a method (its own or from another class) which is deprecated (i.e. which has a variant with a custom generator), then we should also adapt this class to transfer its random generator to this method (by using the variant). However, I did not spot any such class, so it seems there was no interdependency between classes using random generators (I only worked on the core project).

I don't have proof for this process to be correct regarding the goal, so I am working on additional tests. They are not exhaustive, but they provide a safety net. Each of them follows this process:
- decide about a configuration to use,
- listen on the singleton generations to check that this configuration actually uses it,
- execute the legacy version to confirm,
- make a custom generator and listen on it,
- execute the variant with the custom and check it uses the custom + does not use the singleton.

For listening the singleton, I have made a decorator for `PseudoRandomGenerator`. It will come with the next commits. I am currently working on the tests, so they will come together in a separated commit.

ajnebro:
Matthieu, sorry for not paying attentin to this issue before. I'm going to take a look it right now.

matthieu-vergne:
\{^o^}/

ajnebro:
If I have understood well, you have defined to functional interfaces for random number generators, and the idea is to pass the random number generator to all the classes needing it. I assume then that the particular implementation will be injected in a runner program or similar. This certainly allows to remove the Singleton. 

My main concern is how to ensure reproducibility. For example, in class `HUXCrossver` there are two constructors:

```java
public HUXCrossover(double crossoverProbability) {
  this(crossoverProbability, () -> JMetalRandom.getInstance().nextDouble());
}

public HUXCrossover(double crossoverProbability, RandomGenerator<Double> randomGenerator) {
  if (crossoverProbability < 0) {
    throw new JMetalException("Crossover probability is negative: " + crossoverProbability) ;
  }

  this.crossoverProbability = crossoverProbability ;
  this.randomGenerator = randomGenerator ;
}
```
which could cause problems (although after viewing the constructor usages, it can be removed easily).

matthieu-vergne:
The idea of introducing these functional interfaces is basically to apply the interface segregation principle (I in [SOLID](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design))). More precisely, this classes are not supposed to decide what is a good random generator. They focus on how to use it in the right way for their own purpose. For that, they should request a random generator, and should only ask it to fulfil the relevant requirements. Basically, each class should then have its own interface defining exactly the right method. Instead of that, I built on the fact that you use common patterns to design the two main variants (bounded and unbounded) to which I added generics to tell which kind of value to generate. But we could imagine several refinements. I did not do them because it would have an impact on the signatures of of the methods which would not be legacy-friendly. But you can have a glimpse of it with the last commit I made so far.

Now, because their requirements for the random generator stops at this level, reproducibility for the random generator is also not part of their own requirements, but part of the requirements of their user, who should provide the same generator to obtain the same generation. But to understand why it makes sense, one should consider how these classes are used.

Let assume you have an experiment, in which you have several algorithms, which use several operators. Some operators need a random generator (e.g. selection of random individuals, mutation of a random gene), and others not (e.g. selection of best individuals). Some algorithms may also exploit a random generator directly (e.g. random selection of which mutation operator to apply) and others not. Finally, you may want to exploit that at the experiment level too (e.g. run different algorithm settings in a random order to distribute the impact of the instability of the system). What is important to see is that none of them are supposed to know what is a good random generator. All what they need is this generator to provide a distribution of a given kind (e.g. law, bounds). As such, they only ask the caller to provide a concrete generator which satisfies these requirements.

From that point, 2 cases can arise. In both cases, I assume that experiments, algorithms, and operators are independent, such that the dependency inversion principle (D in SOLID) is respected. In other words, you first instantiate the relevant operators, then instantiate the relevant algorithms by giving them the operators, then instantiate the experiment by giving them the algorithms. It might not be exactly that way, we may use for example `Supplier` to instantiate on demand rather than from the start, but the point is that the *choice of which implementation* to use is made before to instantiate the one which will use it. It is not the algorithm which decides which concrete operator to use, neither the experiment which decides which algorithm to execute. The choice is made at a higher level by the user or some other procedure. From that point, the random generators can be given in two ways.

# Immediate Assignment

In this case, you give the random generator to the *constructor*, in other words at the time of the instantiation. Due to that, you first instantiate the operators by giving them the right parameters and the random generators you want to use. Then you instantiate the algorithms by giving them the relevant operators and the random generators to use for themselves. Finally, you instantiate the experiment and give it the relevant random generators if it also requires one. In this case, all the decisions are made at the level where you do all these instantiations, like a runner.

This kind of assignment allows the implementations to assume a given generator for themselves and leave other instances manage their own. So the algorithm is coded to use its own generators and use its operators without caring whether or not they need one. This is how should be designed most of the generic codes (requiring a random generator is rather specific, so it rarely takes place in a generic interface). This is how most of my changes are done, but it was not intended: I just replaced the code in a systematic way, the code was already like that.

# On-Demand Assignment

In this case, the random generators are not requested immediately, but on demand when calling the *methods* using them. In other words, you don't care about which random generator to use at the instantiation time and delay this decision to the last second. It may appear more flexible but, as I mentioned before, neither the operators, the algorithms, or the experiment have the responsibility to decide what is a good generator. In other words, when you dig in the code of an operator, there is no clear reason to find there the instantiation of a specific random generator, which means that this decision will be delegated to the caller, and thus requested as a parameter of the method called. Similarly, the algorithm is not the one responsible of deciding what is a good random generator, which means that it will provide the one it has to the operator when it requests it, but it will also delegate the instantiation decision to its own caller. And again for the experiment. This is what makes the assignment on-demand: after you instantiate everything, you run the experiment by giving it the right random generator. It runs and use it until it has to execute an algorithm which needs it, leading to provide it as a parameter. This algorithm runs and use it if necessary, before to reach the operators requesting it.

This kind of implementation means that the methods called are known enough in details to be aware that a random generator (or several) should be provided. If we make generic code, this is not gonna happen often, but if you implement static methods you cannot store a state, which is why this is the solution I used for `SolutionListUtils.selectNRandomDifferentSolutions(...)`, `SolutionUtils.getBestSolution(...)`. It can also be done for methods which are specific (i.e. not requested by an abstraction) like `AdaptiveGrid.rouletteWheel()`, but I think this kind of design lacks genericness. This is another topic, so let's pass on this point. What is important to understand is that requesting the random generator, and more broadly a parameter, at the method level is adapted to fundamental parameters, while a parameter which is required by specific variants should favour the constructor to let the methods free to implement a more generic shape. I say constructor, but I consider setters as equivalent (you can first instantiate and immediately after use the setter).

# Reproducibility

What is important to understand is that the reproducibility requirement is not a technical detail, but a decisional one. It is about ensuring that a given instance (of a random generator) is the only one to be used. For this requirement to be satisfied, one should focus on where the decision is taken, which is neither at the level of the operator, nor the algorithm, nor the experiment. It is an independent decision, taken depending on the context at the highest level, where we decide what to instantiate. In both the immediate and the on-demand case, because operators, algorithms and experiments are not supposed to know what is a correct implementation for their random generators, they should delegate this decision to their caller, which is why in any case this decision concentrates at the highest level.

Even your singleton does not enforce reproducibility, since you added `setRandomGenerator(...)` to change it on demand. Indeed the "instance" is the same, but this instance is only a wrapper, not the one generating the values, and this setter is what allows you to change the internal generator. Thus, nothing forbid some piece of code somewhere to take the singleton and change its generator. They may have the delicacy to set it back to its old instance when it is done, but if it runs in parallel or if their is other dependencies not spotted because of abstraction, reproducibility is already broken. And this is the whole point: your singleton only gives the illusion of uniqueness without ensuring it because it is not a matter of formal proof (abstraction makes you unable to enforce a specific piece of code to be used or not in a given implementation) but a matter of design choice. This is because it makes no sense to *decide* which implementation to use in a given operator that it won't happen. This is because it makes no sense to *decide* which implementation to use in an algorithm that it won't happen. Your singleton does not (technically) forbid someone to implement an algorithm which uses your singleton here and the basic `Random` somewhere else. You cannot enforce that. All what you can do is give the possibility to reuse easily a generator, such that the programmer can decide at a given level to use an implementation and just use it everywhere if he wants the whole to be reproducible. Which is why it is important to use functional interfaces, to easily map the random generator to the requirements. Having generic ones makes it even easier (as long as they are well done).

In brief, in order to ensure reproducibility, you need first to *assume* that no implementation will instantiate its own generator, but always ask it in argument or through a setter. Second, you need to remove all these legacy methods, to avoid people to use the default generator. We can make them deprecated first to avoid breaking existing code and to support people for the evolution, but the point is to never use them again inside the lib, so the deprecation warnings should help us to spot and replace them.

matthieu-vergne:
A concrete example of how your singleton does not ensure reproducibility
even in your own lib: your class
org.uma.jmetal.solution.impl.DefaultIntegerPermutationSolution uses
java.util.Collections.shuffle(randomSequence), not JMetalRandom.

matthieu-vergne:
Here are the tests. The first commit provides the decorator I use to audit the `JMetalRandom` instance, and the second one the tests themselves. Again, these tests are just a safety net, a partial coverage which only check a single configuration each time to see whether or not using a custom random generator actually replaces the use of the singleton.

ajnebro:
> A concrete example of how your singleton does not ensure reproducibility
even in your own lib: your class org.uma.jmetal.solution.impl.DefaultIntegerPermutationSolution uses java.util.Collections.shuffle(randomSequence), not JMetalRandom.

You are right.

matthieu-vergne:
Yeah, my first pull request which increases significantly the coverage (+1.42%)!

ajnebro:
I have accepted the pull request. What I see then is that you have the choice of use any user defined random generator, but by default the provided singleton implementation is used. This seems to ensure that everything works as before.

ajnebro:
If I have undestood well, now we can write code like this:
```java
    PseudoRandomGenerator pseudoRandomGenerator1 = new JavaRandomGenerator() ;
    pseudoRandomGenerator1.setSeed(0);
    RandomGenerator<Double> randomGenerator1 = () -> pseudoRandomGenerator1.nextDouble() ;

    PseudoRandomGenerator pseudoRandomGenerator2 = new MersenneTwisterGenerator() ;
    pseudoRandomGenerator2.setSeed(0);
    RandomGenerator<Double> randomGenerator2 = () -> pseudoRandomGenerator2.nextDouble() ;
```

ajnebro:
But then, to ensure reproductibility, we should propagate the random generator from the runner code to all the algorithmic components that use it; otherwise, currently reproductibiliby only is guaranteed by using the default singleton-based random generator.

matthieu-vergne:
> I have accepted the pull request. What I see then is that you have the choice of use any user defined random generator, but by default the provided singleton implementation is used. This seems to ensure that everything works as before.

Exact. This is what I try to do for every pull request now: ensure that legacy code still works. If I need to change code in a way which might brake legacy code, I would do it by adding new classes (unless it is an actual fix, but I would ask for it first, like in #206).

> But then, to ensure reproductibility, we should propagate the random generator from the runner code to all the algorithmic components that use it; otherwise, currently reproductibiliby only is guaranteed by using the default singleton-based random generator.

Yes. Which is why in future changes, the point is to keep a list L of the classes which can now manage custom generators, a list L' of classes still using only the singleton, and pay attention to dependencies between the two (especially circular ones if there is) to go ahead. The point is to ensure that a class (i) have the relevant random generator fields instantiated -normally with the singleton- in the constructor, (ii) ensure that these fields are given to the relevant classes instead of using their own default generators, and (iii) add a constructor to customise these fields.

matthieu-vergne:
By the way, do you have a dependency tree of the core project? I don't have anything installed for that, but if you have a good tool I may just exploit what you can get out of it (or install it too).

ajnebro:
No :(

ajnebro:
I like the idea of using functional interfaces, but there is a potential problem in the current approach for random number generators. The `PseudoRandomGenerator` interface has a `nextDouble()` and a `nextInt()` method. Thinking again in reproductibility, if a class needs both methods, then two instances of `RandomGenerator` should be created and propagated until the class using them. 

It's not a big problem, but the issue is that many classes coulnd need to have two random generators as parameters, what it doesn't seem very attractive.

matthieu-vergne:
I don't think it is a problem for 2 reasons.

The first reason is technical: you can generate an int from a double by removing the decimal part. More generally, if you need to generate several kinds of random data, you can always ask a [0, 1[ random generator and create other random generators based on it. We can add factory methods to produce easily generators using other generators to make it simple, and nothing forbid the algorithm to have a detailed constructor which requires several random generators + another constructor with a single, generic generator used for everything (by using these factory methods).

The second reason is conceptual : if you need both an int and a double, it means that you need to generate 2 kinds of data, thus it makes sense to have has much suppliers. Whether or not you use the same instance for all of them is not a decision of the algorithm itself, but of the user. The algorithm does not has the responsibility of deciding what is a correct generator, only how to use it.

PR #200, opened 2017-04-14
MOEAD-DD added to jMetal 5

vinixnan:
I just adapted MOEAD-DD code from @JerryI00 (available in http://www.cs.bham.ac.uk/~likw/code/MOEADD.zip)  to jMetal 5.

To perform it:
- I copied and paste his code
- Created some hashmaps to avoid the need of Solution Class changes.
- make the class extends AbstractMoead class
- changed Vectors to ArrayList
- changed some ifs to Switch
- removed some unnecessary code

I do not know how I could properly test it.I made some tests using WFG1 and it worked well. However, I ask for help in this task.

@JerryI00 @ajnebro @fritsche Could you help to test and validate this good algorithm?


JerryI00:
Thanks a lot for @vinixnan effort on this transform. It looks fine to me.

PR #201, opened 2017-04-17
Refactor ChartContainer. Add NSGAIIMeasuresWithChartRunner.

georgero5:
Changes to tidy up the ChartContainer class and adding support for adding Indicators to the chart.
Added NSGAIIMeasuresWithChartsRunner class that uses the ChartContainer to display an Hypervolume chart.


matthieu-vergne:
Please preserve the style of the code. It is hard to evaluate the actual
changes of a commit when they are mixed with irrelevant changes in the
indentation and other spacing modifications.

georgero5:
Indentation changes made by eclipse's auto-format have been reverted to preserve the style of the code.
The constructors for DMOPSO and DMOPSOMeasure classes have been refactored.

matthieu-vergne:
Thank you for your effort. {^_^}

I am not familiar with these algorithms, so I only comment on qualitative
aspects.

It seems to me that ChartContainer.SetFrontChart(objective1, objective2,
referenceFront) mixes stuff. In particular, the reference front seems to be
used only for display if it is not null. Looks to me like something which
should be in a different method and has nothing to do with the setter.
Additionally, it is not the reference front, but its file: if we follow the
method calls, at some point the reference front becomes a file name and we
build the actual front from it. This could also be clarified.

In UpdateFrontCharts(solutionList), you create xData and yData variables
before to use them in two cases, but you should avoid this kind of C-like
coding if it is not required. You should create variables when you
instantiate them, and split only if you actually need to increase their
scope. Here, their use with the front chart and their use with the var
chart are independent, while you allow a reuse by increasing their scope to
the whole method. If their reuse is not wanted, you should instantiate them
twice. With the current code, you may remove some code in the var chart
case without having a compilation error, and figure it out at runtime only.

In getSolutionsForObjective(solutionList, objective) and getVariableValues(
solutionList, variable) you may use parallel streams since each iteration
is independent. Depending on the implementation, browsing sequentially
through solutionList.get(i) might be rather inefficient.

In DMOPSOMeasuresRunner, you instantiate a DMOPSOMeasures, so it is a pity
to store it in a DMOPSO variable and cast it everytime. There is the same
issue with using builders, but that is a more global issue.

Issue #71, opened 2015-05-20
Problem-solution spaghetti code


matthieu-vergne:
I am currently looking for fixing the warnings related to missing parameters for generics and I see that you have spaghetti code between problems and solutions implementations.

A solution is something with rather many properties:
```
public interface Solution<T> extends Serializable {
  public void setObjective(int index, double value) ;
  public double getObjective(int index) ;

  public T getVariableValue(int index) ;
  public void setVariableValue(int index, T value) ;
  public String getVariableValueString(int index) ;

  public int getNumberOfVariables() ;
  public int getNumberOfObjectives() ;

  public double getOverallConstraintViolationDegree() ;
  public void setOverallConstraintViolationDegree(double violationDegree) ;

  public int getNumberOfViolatedConstraints() ;
  public void setNumberOfViolatedConstraints(int numberOfViolatedConstraints) ;

  public Solution<T> copy() ;

  public void setAttribute(Object id, Object value) ;
  public Object getAttribute(Object id) ;
}
```
Why not. I think there is a lot to remove but let that for later.

The point is that you have provided a generic implementation through `AbstractGenericSolution`, and here is the issue:
```
public abstract class AbstractGenericSolution<T, P extends Problem> implements Solution<T> {
  ...
}
```
Thus, the solution depends on the problem. It could seem natural on an informal basis, but if we look at the definition of a problem:
```
public interface Problem<S extends Solution> extends Serializable {
  /* Getters */
  public int getNumberOfVariables() ;
  public int getNumberOfObjectives() ;
  public int getNumberOfConstraints() ;
  public String getName() ;

  /* Methods */
  public void evaluate(S solution) ;
  public S createSolution() ;
}
```
So here you see that a problem also depends on solutions, thus a loop. Loops are hints that responsibilities are not well delegated, because to do its job the first needs the second which, to do its own job, needs the first. As an additional hint, you can see that in both problems and solutions you have the number of variables and objectives, and while you have the number of violated constraints in the solution you don't have the number of constraints, but you find it in the problem. For me, this shows a clear lack of separation of responsibilities, and it should be fixed. Because while I was trying to figure out what happen in the implementations to fix the missing generics, it was just confusing and exploding everywhere and I actually had to change the code itself, but it is explosive.

[...]

PR #205, opened 2017-04-28
Generalisation

matthieu-vergne:
Here I remove unused dependencies to `Solution<T>`, so the classes can be used in more contexts. In the process, I have noticed some classes that makes such a generalisation harder, like `GenericSolutionAttribute` and others. Future generalisations will probably focus on them.

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/coralreefsoptimization/CoralReefsOptimization.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/coralreefsoptimization/CoralReefsOptimization.java
@@ -5,7 +5,6 @@ import org.uma.jmetal.operator.CrossoverOperator;
 import org.uma.jmetal.operator.MutationOperator;
 import org.uma.jmetal.operator.SelectionOperator;
 import org.uma.jmetal.problem.Problem;
-import org.uma.jmetal.solution.Solution;
 import org.uma.jmetal.util.pseudorandom.impl.MersenneTwisterGenerator;
 
 import java.util.ArrayList;
@@ -18,7 +17,7 @@ import java.util.List;
  * @author Inacio Medeiros <inaciogmedeiros@gmail.com>
  * 
  */
-public class CoralReefsOptimization<S extends Solution<?>>
+public class CoralReefsOptimization<S>
 		extends AbstractCoralReefsOptimization<S, List<S>> {
 
 	private Problem<S> problem;
--- a/jmetal-core/src/main/java/org/uma/jmetal/algorithm/impl/AbstractCoralReefsOptimization.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/algorithm/impl/AbstractCoralReefsOptimization.java
@@ -9,7 +9,6 @@ import org.uma.jmetal.algorithm.Algorithm;
 import org.uma.jmetal.operator.CrossoverOperator;
 import org.uma.jmetal.operator.MutationOperator;
 import org.uma.jmetal.operator.SelectionOperator;
-import org.uma.jmetal.solution.Solution;
 
 /**
  * Abstract class representing a Coral Reefs Optimization Algorithm
@@ -23,7 +22,7 @@ import org.uma.jmetal.solution.Solution;
  * @author Inacio Medeiros <inaciogmedeiros@gmail.com>
  */
 @SuppressWarnings("serial")
-public abstract class AbstractCoralReefsOptimization<S extends Solution<?>, R>
+public abstract class AbstractCoralReefsOptimization<S, R>
 		implements Algorithm<R> {
 
 	protected List<S> population;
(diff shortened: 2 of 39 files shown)


ajnebro:
Hmmm, the point is that all those classes will use subclasses of Solution<T>,  hence the inclusion of the dependence.

matthieu-vergne:
These classes do not, thus the removal. The subclasses need it, so the subclasses introduce the dependence, but placing this dependence at a higher level because the lower level use it is a mistake. It only reduces the range of use of the generic classes, without adding value. If at some point you use `Solution` specific features in the high level classes, then introduce the dependence, but not otherwise.

matthieu-vergne:
My plan is to progressively make your algorithms more generic in order to be more flexible on the kind of solutions they can manage, thus making them usable directly on solution implementations made by the users of the lib rather than being restricted to use JMetal-specific implementations. In particular solutions imported from other libs on which they might not have the ability to adapt the code to fit your current algorithms. Such a generalisation should be achievable without decreasing performance.

ajnebro:
Well, I personally don't see any inconvenience of that but, thinking in many potential jMetal users, finding related classes and some of them including `S extends Solution<?>` and others only with `S` can be confusing.

matthieu-vergne:
Why should it be confusing? Are you confused when you use a `List<T>` to store solutions and other stuff? I agree that having a code easy to understand is important, but this is your habit which speaks here. Implementations depending on `Solution` can be used with it, those not depending on it, like `List<T>`, can be used as well. There is nothing to be confused here.

matthieu-vergne:
What I do here is exactly the same than [one of the things you agreed with me](https://github.com/jMetal/jMetal/issues/71#issuecomment-105541077) in issue #71.

matthieu-vergne:
By the way, some classes genuinely depend on your solution interface, so they cannot be generalised without breaking their method signatures, because generalising them requires to ask more information from the user. I don't plan to change these ones, in order to remain compatible with legacy code.

matthieu-vergne:
To be more precise, there is basically 3 levels:
- the generic interfaces
- the intermediary classes, which help implement typical algorithms
- the concrete implementations

The highest level, which is the root of the framework, is done. If there is changes to be made, they should be really well motivated. The concrete implementations are also done, in the sense that they are the one genuinely exploiting the specificities of `Solution` to run. I don't plan to change them either, so the basic users (who do not implement new algorithms, only use existing ones) should not be impacted at all. I work at the intermediary level, in order to ensure that the intermediary classes, which should help people to create easily new algorithms, are not over constrained but still provide the relevant constraints to reduces the mistakes.

matthieu-vergne:
What happened?

ajnebro:
Many compiling errors after accepting the pull request about Generalization. I have reverted the pull request for now (the first time I do something similar).

ajnebro:
I'm affraid I did something wrong ...

matthieu-vergne:
Need help?

ajnebro:
The fact is that I accepted first pull request  #208 and then #205. After accepting the second one some errors appeared.

Certainly, I'm not sure about how to proceed now, because I don't know if reverting pull request #208 and accepting both in the right order would work.

matthieu-vergne:
If there is a merging problem, I doubt the order changes anything. I will cherry pick generalisation, fix what seems to be wrong, and make another pull request.

ajnebro:
Ok. Thanks :)

matthieu-vergne:
I checked the merging, but there was only 1 error, because one of the methods added to customize the random generators (`selectNRandomDifferentSolutions(..., randomGenerator)`) was not covered by the generalisation. The fix was only to remove the `extends Solution<T>` of this additional method. I just run all the tests without problem.

I will make a new pull request with this addition.

ajnebro:
By the way, I was thinking in closing version 5.3 after your contributions with the random generators and the graphics code of Jorge.

matthieu-vergne:
OK. Do you speak about the complete revision of random generators? Or only what has been done so far?

ajnebro:
Thanks for checking the merging :)

ajnebro:
All the changes and additions since the last release. I think that there are enough stuff.

The "worst" part is documentation ... I'm not particularly happy with using merely MarkDown code. It's simple, but sometimes I miss using ... Latex.

matthieu-vergne:
OK, so I have time for the remaining stuff. That sounds good.

If you think I could add something to the doc, especially about the stuff I made, feel free to tell me where.

ajnebro:
Thanks :)

I will wait until you tell me whether you have finished with the random generators.

matthieu-vergne:
Uh? Not sure to understand.

**So far:** new functional interfaces + factory methods + refine the leaf classes which use `JMetalRandom` by adding the possibility to use a custom one.

**Not done yet:** refine the classes which use the leaf ones and further (e.g. algorithms). These ones may be subject to cyclic dependencies, so a particular care is needed. It might take some time to be done.

Do you plan to close 5.3 with what is done **so far** or do you plan to wait for what is **not done yet** before to close it? For me, you can close it now, because although the support for customization is not full, it is still working where it has been applied, so new code can exploit it.

ajnebro:
> plan to wait for what is not done yet before to close it

;)

You are spending a lot time in the random generator issue, and I prefer to wait until you finish. In the meantime, I will document the new chart stuff.

matthieu-vergne:
OK. I'll try to get it done as soon as possible.

matthieu-vergne:
Well, I can try to go fast, but it will also depend on you replying to my related questions. {'^_^}

Issue #212 is first of all a clarification/confirmation of the interpretation (I should be able to adapt my changes if I am wrong), so it should not take long to close.

Issue #211 is more tricky, because it relates to a design choice. Technically speaking, it is not blocking (I can just add stuff and ignore that point), but then we have redundant stuff which makes it more complex that required and threatening for the reproducibility requirement. So I think it is important to not ignore it.

PR #207, opened 2017-05-07
Added support for adding a Reference Point to a chart in the ChartContainer

georgero5:
- Fixed a "bug" in MOEAD, population should be reset in the initializePopulation so it is reset when running several independent runs of MOEAD in a experiment.
- A Reference Point can be added to the Front chart, see example in WASFGAMeasuresRunner.
- Tidy up ChartContainer code.


matthieu-vergne:
Some tests to ensure that future changes make no regression would be nice,
if you can.

2017-05-08 5:23 GMT+09:00 Codecov <notifications@github.com>:

PR #213, opened 2017-05-11
Fix warnings.

matthieu-vergne:


matthieu-vergne:
Sorry, I included stuff already in #203. I'll remove them later today.

PR #215, opened 2017-05-27
Generify PointCrossover and add support for NPointCrossovers

FlapKap:
Right now SinglePointCrossover only works with BinarySolutions. This aims to ultimately make it work with all solutions, though I haven't changed SinglePointCrossover since that and it's test seems somewhat confusing to me, and this is my first contribution to any OSS project.


ajnebro:
Hi @FlapKap.
  Thanks for your contribution. 

  Let me check your updates before accepting them.



Antonio

matthieu-vergne:
Generalization? That sounds interesting. I might look at it later (after
acceptance if I cannot do it before).

PR #216, opened 2017-07-25
Added two classes (Task.java and Machine.java) as the variables for

shambakey1:
scheduling problem

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/util/Machine.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/util/Machine.java
@@ -1,0 +1,827 @@
+package jmetal.util;
+
+import java.util.Collections;
+import java.util.Vector;
+import java.util.Comparator;
+import java.util.Random;
+import java.util.Iterator;
+
+/**
+ * This class represents computing elements such as machines, processors, nodes.
+ * It is useful for scheduling problems like jobshop and flowshop
+ * 
+ * @author shambakey1
+ */
+public class Machine {
+	/*
+	 * Original members that describes this machine
+	 */
+	private int id; // Machine identifier
+	private int db_id; // Machine identifier recorded in database_conf table
+	private long speed; // Speed of current machine
+	private Vector<Task> ass_tasks; // Assigned tasks to this machine
+
+	/*
+	 * Metric members that are used to measure performance of this machine
+	 * (i.e., span)
+	 */
+	private double span; // Final span of this machine due to assigned tasks.
+
+	/**
+	 * Default constructor. ID, speed and span are set to 0 which is initial
+	 * invalid value.
+	 */
+	public Machine() {
+		this.id = 0;
+		this.speed = 0;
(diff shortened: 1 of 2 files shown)


matthieu-vergne:
Please test your code before to push it: `DBInOut` is missing, so it cannot compile. Nevertheless, I looked at the code, and here is my opinion.

**Summary**

These 2 classes are not used in jMetal, and none of them use jMetal stuff, which for me is enough to say that they have nothing to do here.

**Details**

These classes seem way too specific to me to be a requirement for metaheuristics in general. If they are useful for scheduling problems, you may think about adding them to the module `jmetal-problem`, instead of `jmetal-core`, but I would personally reject them if no clear evaluation of their added value is done. If it appears to be something that we do not use internally or that we cannot show to be a significant help for jMetal users in general, then there is no point adding them. Anything *could* be useful, as long as 1 person uses it, but if we look at jMetal as a library, we cannot afford to implement any specific stuff which *could* be useful. We have to remain generic enough to cover many cases, and not overwhelm ourselves in maintaining code that almost nobody uses. Highly specific code must be implemented by highly specific projects.

Would it be accepted, I have many other comments on the code itself:
- Add unitary tests!!!
- Decouple `Machine` and `Tasks` from the database
- Decouple them from the standard `Random` (e.g. line 171 of `Task`), especially to use our `JMetalRandom`. I recommend to use `Supplier`.
- Remove spaghetti code by properly assigning responsibilities (e.g. a `Machine` stores a set of assigned `Task` instances **or** a `Task` stores the `Machine` it is managed by, but not both)
- Replace old fashioned C code by actual Java code:
	- Use the right convention (e.g. `tmp_machine` to `tmpMachine`).
	- Declare variables at the right scope, not everything at the beginning.
- Replace old fashioned C code by more modern one:
	- Replace small classes by lambda expressions (used directly in the code if they are not used that often), like your machine comparators (possibly with a single `MachineComparator` class which centralises the use of the `signum`).
	- Replace exits by proper, explicit exceptions (e.g. lines 688-691 of `Machine`, lines 90-93 of `Task`).
	- Replace the use of `Vector` by `Set`, because you speak about sets (or `Collection` if the uniqueness is not actually a requirement).
	- Use more explicit namings (e.g. `calc_span` to `spanPerMachine`)
	- Remove excessive comments to favor clear code with explicit variables and methods.
- Revise the code to remove useless stuff (e.g. empty lines whith only a semicolon, redundant code)
- Fix conventions mistakes (e.g. `preSpantaskAssignment` to `preSpanTaskAssignment`) 
- Probably other stuff I missed in my quick reading.

I would finish by saying that it looks a lot like utilitary classes for thread management... but then I am convinced there is way more efficient tools to use for that, and I wonder why it is not related to standard classes like `Thread`, `Future`, or `Executor`. The notions of execution time and speed used in these classes seem to me particularly naive and unreliable. Of course, this comment is relevant only if these classes aim to support thread management.

Issue #223, opened 2017-08-26
NSGAIII EnviromentalSelection is removing multiple solutions per iteration when finding a niche reference point


Barbiero:
Hiya

I'm currently using Jmetal on a project due to its NSGAIII algorithm, but I was surprised when it crashed on me with out-of-bounds exceptions when I ran my Problem, even though it worked perfectly fine on NSGAII.

After some investigation, I found that, when finding niche reference points at `EnvironmentalSelection#execute`, the iteration expects to remove one member per cycle:
```java
        while (source.size() < this.solutionsToSelect) {
            int min_rp = FindNicheReferencePoint();

            S chosen = SelectClusterMember(this.referencePoints.get(min_rp));
            if (chosen == null) // no potential member in Fl, disregard this reference point
            {
                this.referencePoints.remove(min_rp);
            } else {
                this.referencePoints.get(min_rp).AddMember();
                this.referencePoints.get(min_rp).RemovePotentialMember(chosen);
                source.add(chosen);
            }
        }
```

However, when looking at `ReferencePoint#RemovePotentialMember`, we see that the method is designed to remove multiple, equivalent, members when called:
```java
    public void RemovePotentialMember(S solution) {
      Iterator<Pair<S, Double>> it = this.potentialMembers.iterator();
      while (it.hasNext()) {
        if (it.next().getLeft() == solution) {
          it.remove();
        }
      }
    }
```

this means that, in the event that a `solution` might exist in duplicates(ie. two equivalent solutions might be generated if operators are too conservative), the method will end up removing multiple potential members out of the reference points, but will add only one member - eventually clearing out all reference points while not filling `source` with `solutionsToSelect` solutions, leading to `FindNicheReferencePoint` breaking trying to index an empty array.

I fixed this by simpling `break`ing the method, and now the algorithm is working properly for me; However, I'm not creating a pull request because I'm inexperienced and my approach might be wrong. Should it even be possible for duplicate `solution`s to exist within a `potentialMembers` list in the first place? Or maybe we could simply add a copy of the `chosen` object to `source` for each removed member?

The original paper's algorithm(section IV-E, Algorithm 4) is written in a way that it only removes a single point per iteration, so I feel like my approach is the correct - but I'd prefer to discuss this with more experienced people in the area before considering a PR for this particular issue.

Thank you for your time!

matthieu-vergne:
[...]

PR #225, opened 2017-08-30
Fix NSGAIII-ReferencePoint so it doesn't remove more than one potential member per iteration

Barbiero:
as discussed on issue #223

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/nsgaiii/util/ReferencePoint.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/nsgaiii/util/ReferencePoint.java
@@ -1,13 +1,11 @@
 package org.uma.jmetal.algorithm.multiobjective.nsgaiii.util;
 
-import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.uma.jmetal.solution.Solution;
+import org.uma.jmetal.util.JMetalException;
 import org.uma.jmetal.util.pseudorandom.JMetalRandom;
 
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
 
 /**
  * Created by ajnebro on 5/11/14.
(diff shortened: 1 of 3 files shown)


jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/nsgaiii/util/ReferencePoint.java
```diff
@@ -101,6 +101,7 @@ public void RemovePotentialMember(S solution) {
     while (it.hasNext()) {
       if (it.next().getLeft() == solution) {
```
matthieu-vergne: I notice that it does not remove all the equivalent solutions, but only a single instance (`==` is used, not `.equals(solution)`). Only if the same instance is used several times, then yes it is removed several times. Thus, I would suggest to use `equals` instead of `==`, just in case a different instance happens to be used. It would be an additional fix.

Barbiero: While that's true, in `EnvironmentalSelection#execute` we have this block:
```java
{
	this.referencePoints.get(min_rp).AddMember();
	this.referencePoints.get(min_rp).RemovePotentialMember(chosen);
	source.add(chosen);
}
```
So we could potentially be removing a certain instance from ReferencePoint while adding a different one to `source`, so I'm not sure if that's the correct approach here.

matthieu-vergne: I don't know, I am not familiar with the algorithm. But situations which require to put several times the same instance in a collection, despite having other instances, is not common. There is, like thread reuse, but because it is rare I think it makes the question relevant. At least, if it has been designed with the idea of having a container with unique instances, then the initial code is correct, although performance could be improved with your fix.

Barbiero: I'm a bit concerned changing semantics with this and breaking something unrelated unknowingly; so I'll run a few experiments and report back if it breaks anything.

Barbiero: Doesn't seem like it makes a difference, after running a few multi-threaded experiments. I'm switching it to equals

matthieu-vergne:
Well, it was more a comment for the others, actually. They may know better whether or not such a change makes sense compared to the original algorithm. I did not read the papers, personally. Too lazy for that for now {'^_^}.

jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/nsgaiii/util/ReferencePoint.java
matthieu-vergne: Well, it was more a comment for the others, actually. They may know better whether or not such a change makes sense compared to the original algorithm. I did not read the papers, personally. Too lazy for that for now {'^_^}.

Issue #221, opened 2017-08-18
God object anti-pattern in experiments


matthieu-vergne:
I am currently working on an experiment framework, so I am looking at the current code to see how I can help to improve it. And here is an anti-pattern that I could spot.

A [God object](https://en.wikipedia.org/wiki/God_object) is an object that knows or do many stuff, and thus on which many other objects will rely for their own reasons. Usually, a good factoring consists in finding common stuff and factor it into a dedicated class. The God object occurs when *different* features are centralized into a single class, thus making the responsibility unclear and tentacular.

This is what I can observe with `org.uma.jmetal.util.experiment.Experiment`. First, despite its name, it is not an experiment (i.e. something which produces an evaluation of some algorithm performances) but a configuration container. It stores many things, like the name of the experiment, the list of algorithms to run, the directory in which the results will be stored, the number of runs to perform on each algorithm, the number of cores to use, etc. Consequently, it is related to many responsibilities, like display, storage, execution, etc. It is still a *soft* God object, because it does nothing (the only behavioral method is the one which removes duplicate algorithms), but it is one.

Some clear disadvantages are:
- one must guess what to put in `Experiment`, because the code does not show which parameters are required for each step, as they all request only the God object.
- because of that, whether or not the right parameters are provided is checked at runtime, while with explicit parameters such a control would be made at compilation time.

A better way to go for now is to extract all the data from this God object and make relevant parameters for each class currently using it (`ExecuteAlgorithms`, `ComputeQualityIndicators`, etc.). This way, when you use them, you see clearly what is needed in the code.

If default parameters need to be used, one can simply make some static values somewhere, like an `ExperimentDefault` structure. It would be a temporary solution, because I think there is a lot of improvements which can be done regarding the responsibilities of each step, in order to remove some redundancies.

So what I suggest is to plan its removal (i.e. deprecate `Experiment` and related stuff, and remove its use in jMetal). I am currently working on that in a dedicated branch, so I can make a pull request for it. But first I would like to know your opinion about it.

matthieu-vergne:
For example, here is what you have:
```java
String experimentBaseDirectory = args[0];

List<ExperimentProblem<BinarySolution>> problemList = new ArrayList<>();
problemList.add(new ExperimentProblem<>(new ZDT5()));
problemList.add(new ExperimentProblem<>(new OneZeroMax(512)));

List<ExperimentAlgorithm<BinarySolution, List<BinarySolution>>> algorithmList =
[...]

PR #227, opened 2017-10-07
New Experiment Component

LucasLP:
This component generate a line chart of quality indicator of evolution process 
For example: https://github.com/LucasLP/ExecuteJMetal/blob/master/Additional/example_UF1_HV.png

      The line is the means value of the runs, and dotted lines is the first and third quartiles
  Default behavior: Execute the indicator at each 1% of evaluations;
  Use "HV" "IGD" "Spread" "Epsilon"
   Attention! It can be very slow with HV
  For use this you need:
  **1. (at Algorithm) Implement "HistoricAlgorithm.java" in you algorithm**
  
  **2. (at Algorithm) Add "HistoricData.java" how an attribute and implement methods**
       for example: "Map<String, HistoryData> history;" Name of indicator and data
       @Override
        public HistoryData getHistory(String indicator) {
            return history.get(indicator);
        }
 
        @Override
        public void setRunNumber(Integer runsNumber) {
            history = new HashMap<>();
            history.put("HV", new HistoryData(runsNumber));
            history.put("IGD", new HistoryData(runsNumber));
            history.put("Epsilon", new HistoryData(runsNumber));
            history.put("Spread", new HistoryData(runsNumber)); 
        }
  
  **3. (at Algorithm) Execute indicators**
       for example:
           if(HistoricAlgorithm.testToCalculate(evaluations,maxEvaluations)){
                HistoricAlgorithm.calculateIndicators(evaluations, maxEvaluations, problem.getName(), population,history);
            }
  
  **4. (at Experiment) Generate data files and R scripts**
       for example:
           HistoryData.printAllDataInstances(experiment, indicators);
           new GenerateEvolutionChart(experiment, indicators).run();

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoricAlgorithm.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoricAlgorithm.java
@@ -1,0 +1,41 @@
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package org.uma.jmetal.util.experiment.component.EvolutionChart;
+
+import java.util.List;
+import java.util.Map;
+import org.uma.jmetal.solution.DoubleSolution;
+
+/**
+ *
+ * @author Lucas Prestes <lucas.prestes.lp@gmail.com> 
+ */
+public interface HistoricAlgorithm {
+    public HistoryData getHistory(String indicator);
+    public void setRunNumber(Integer runsNumber);
+    
+    
+    
+    public static boolean testToCalculate(Integer evaluations, Integer maxEvaluations){
+        return (evaluations%(maxEvaluations/100)==0); 
+    }
+    
+    public static void calculateIndicators(Integer evaluations, Integer maxEvaluations, String problemName, List<DoubleSolution> population, Map<String, HistoryData> history) {
+        int timeIndex = (int) (evaluations / (maxEvaluations / 100)) - 1;
+        Map<String, Double> indicators = HistoryData.calculateQualityIndicator(population, problemName, history.keySet());
+        for (String str : history.keySet()) {//for each indicator in history
+            history.get(str).addData(indicators.get(str), timeIndex);//add data
+        }
+    }
+}
--- a/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoryData.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoryData.java
@@ -1,0 +1,162 @@
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package org.uma.jmetal.util.experiment.component.EvolutionChart;
+
+import org.uma.jmetal.util.experiment.component.GenerateEvolutionChart;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.uma.jmetal.qualityindicator.impl.Epsilon;
+import org.uma.jmetal.qualityindicator.impl.InvertedGenerationalDistancePlus;
+import org.uma.jmetal.qualityindicator.impl.Spread;
+import org.uma.jmetal.qualityindicator.impl.hypervolume.PISAHypervolume;
+import org.uma.jmetal.solution.DoubleSolution;
+import org.uma.jmetal.util.experiment.Experiment;
+import org.uma.jmetal.util.experiment.util.ExperimentAlgorithm;
+import org.uma.jmetal.util.front.Front;
+import org.uma.jmetal.util.front.imp.ArrayFront;
+import org.uma.jmetal.util.front.util.FrontNormalizer;
+import org.uma.jmetal.util.front.util.FrontUtils;
+import org.uma.jmetal.util.point.util.PointSolution;
+
+/**
+ *
+ * @author Lucas Prestes <lucas.prestes.lp@gmail.com> 
+ */
+public class HistoryData {
+    public static final String DEFAULT_BASE = "history";
+    private Integer testId;
+    
+    private Integer numberOfData;
+    private Integer numberOfTest;
+    
+    /**
+     * Number of data (%) / number of test
+     */
+    private Double[][] history;
+   
+
+    public HistoryData(Integer numberOfTest) {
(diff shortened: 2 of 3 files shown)


jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoricAlgorithm.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.List;
+import java.util.Map;
+import org.uma.jmetal.solution.DoubleSolution;
+
+/**
+ *
+ * @author Lucas Prestes <lucas.prestes.lp@gmail.com> 
+ */
+public interface HistoricAlgorithm {
```
matthieu-vergne: I don't get the purpose of having an algorithm interface to make an experiment tool available. We must ensure that dependencies restrict to the minimum, so we cannot impose to an algorithm to implement a specific interface for the sole purpose of using one experiment component (this is what `HistoricAlgorithm` makes me suppose with its naming). Rather, the component should be independent, and it should be able to receive data whatever the source, so we can choose for instance to feed it directly with the outputs of a running algorithm (or with its measures) or feed it with already stored data. Can you imagine if we implement an interface for each component we want to experiment? Basically, a good experiment component should be algorithm-agnostic.

jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoryData.java
```diff
@@ -0,0 +1,162 @@
...
+
+    public HistoryData(Integer numberOfTest) {
+        this.testId = -1;
+        this.numberOfData = 100;
+        this.numberOfTest = numberOfTest;
+        history = new Double[this.numberOfData][this.numberOfTest];
+
+        for (int i = 0; i < this.numberOfData; i++) {
+            for (int j = 0; j < this.numberOfTest; j++) {
+                history[i][j] = 0.0;
```
matthieu-vergne: If you never have null values, you may instead use  the primitive type `double` instead of its object counterpart. It would also allow you to avoid the initialisation phase (a `double` is by default at `0.0`).

jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoryData.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+        
+    }
+    
+    public void addData(Double data, Integer timeIndex){
+        if(timeIndex==0)testId++;
+        history[timeIndex][testId] = data;
```
matthieu-vergne: Missing indexes check (`IndexOutOfBoundsException`).

jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoryData.java
```diff
@@ -0,0 +1,162 @@
...
+                for (int data = 0; data < numberOfData; data++) {
+                    ps.print(history[data][test]);
+                    if(data<numberOfData-1){
+                        ps.print("\t");
+                    }
+                }
+                ps.print("\n");
+            }
+        } catch (IOException ex) {
+            Logger.getLogger(GenerateEvolutionChart.class.getName()).log(Level.SEVERE, null, ex);
```
matthieu-vergne: I would say wrong class in the logger. And I would suggest to declare the logger as a static member of the class, so we can reuse it anywhere in the class.

jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoryData.java
```diff
@@ -0,0 +1,162 @@
...
+            for (String string : indicatorToSave) {
+                if(string.equals("HV")){
+                    indicators.put("HV", new PISAHypervolume<PointSolution>(normalizedReferenceFront).evaluate(normalizedPopulation));
+                } else if(string.equals("Epsilon")){
+                    indicators.put("Epsilon", new Epsilon<PointSolution>(normalizedReferenceFront).evaluate(normalizedPopulation));
+                } else if(string.equals("IGD")){
+                    indicators.put("IGD", new InvertedGenerationalDistancePlus<PointSolution>(normalizedReferenceFront).evaluate(normalizedPopulation));
+                } else if(string.equals("Spread")){
+                    indicators.put("Spread", new Spread<PointSolution>(normalizedReferenceFront).evaluate(normalizedPopulation));
+                }
```
matthieu-vergne: Rather than strings, an `enum` seems relevant to me here. Moreover, that suite of conditions would be better factored as a switch-method, like:
```java
private static enum Indicator {HV, Epsilon, IGD, Spread}

private Double getIndicatorValue(Indicator indicator) {
	switch indicator {
		case HV: return new PISAHypervolume<PointSolution>(normalizedReferenceFront).evaluate(normalizedPopulation);
		case Epsilon: return new Epsilon<PointSolution>(normalizedReferenceFront).evaluate(normalizedPopulation);
		case IGD: return new InvertedGenerationalDistancePlus<PointSolution>(normalizedReferenceFront).evaluate(normalizedPopulation);
		case Spread: return new Spread<PointSolution>(normalizedReferenceFront).evaluate(normalizedPopulation);
		default: throw new IllegalArgumentException("Unmanaged indicator: " + indicator);
	}
}
```
It would reduce the loop to:
```java
for (String indicator : indicatorToSave) {
	indicators.put(indicator, getIndicatorValue(indicator));
}
```
Which can further be refined through lambdas (just for curiosity, there is no real added value here):
```java
indicatorToSave.stream().peek(indicator -> indicators.put(indicator, getIndicatorValue(indicator)))
```

jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoryData.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+            
+        } catch (FileNotFoundException ex) { 
+            Logger.getLogger(GenerateEvolutionChart.class.getName()).log(Level.SEVERE, null, ex);
```
matthieu-vergne: Wrong class?

jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/EvolutionChart/HistoryData.java
```diff
@@ -0,0 +1,162 @@
...
+        return indicators;
+    }
+    
+    
+    
+    public static void printAllDataInstances(Experiment<?,?> experiment, List<String> indicators){
+        System.out.println("Generating data files of algorithms evolution...");
+        for (ExperimentAlgorithm<?, ?> experimentAlgorithm : experiment.getAlgorithmList()) {  
+            for (String indicator : indicators) {
+                HistoryData hd = ((HistoricAlgorithm) experimentAlgorithm.getAlgorithm()).getHistory(indicator);
```
matthieu-vergne: Rather than taking the whole experiment, you should take only the relevant data, like `Collection<Map<Indicator, HistoryData>>`. If you need other stuff stored in the experiment structure, you should just ignore the structure and request what you need in argument. You don't need the whole experiment, only some relevant pieces, so don't depend on the whole stuff. We are in a single experiment tool, here, its scope should not be as broad as a whole experiment. `Experiment` should not be a God object, as explained in #221.

jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/GenerateEvolutionChart.java
```diff
@@ -0,0 +1,199 @@
...
+ *      for example:
+ *          HistoryData.printAllDataInstances(experiment, indicators);
+ *          new GenerateEvolutionChart(experiment, indicators).run();
+ *          
+ * 
+ * @author Lucas Prestes <lucas.prestes.lp@gmail.com> 
+ */
+public class GenerateEvolutionChart implements ExperimentComponent{
+    
+    Experiment<?,?> experiment;
```
matthieu-vergne: An experiment component depending on a whole experiment... something wrong here. Not yet spaghetti code, but close to be.

jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/GenerateEvolutionChart.java
```diff
@@ -0,0 +1,199 @@
...
+    List<String> indicators;
+    String outputDir;
+    
+    public GenerateEvolutionChart(Experiment experiment, List<String> indicators) {
```
matthieu-vergne: Don't forget generics.

jmetal-core/src/main/java/org/uma/jmetal/util/experiment/component/GenerateEvolutionChart.java
```diff
@@ -0,0 +1,199 @@
...
+            }
+        }
+    }
+    
+    
+    
+    
+    private String generateRscript(List<ExperimentAlgorithm<DoubleSolution, List<DoubleSolution>> > algorithmList, String indicator, String problem){
+        int i = 1;
+        String str = "# Read data file\n" ;
```
matthieu-vergne: Use the `PrintStream` here rather than string concatenation. Concatenation is costly, and `PrintStream` allows you to use `println` to avoid expliciting newlines.

PR #231, opened 2017-10-30
Changed case of class names in CEC benchmark array

leechristie:
CEC benchmark class names were changed to PascalCase by bc7f1ea129c6f81b1a1799fe1c662acf11538ca1. Benchmark currently retains old case.


matthieu-vergne:
Given they are strings, it does not seem that relevant to me, as opposed to renaming variables and alike...

leechristie:
On line 959 the strings are used to instantiate via reflection: `loader.loadClass(CEC2005Code + "." + test_func_class_names[func_num - 1])` so it will throw `ClassNotFoundException` if the string case doesn't match the class name.

matthieu-vergne:
Then it makes much more sense. I thought it was mere Java naming convention.

PR #233, opened 2017-11-14
Replace HashMap by solution attributes

fritsche:
I have experienced OutOfMemory using MOEA/DD for many objective problems
I noticed that the HashMaps were growing during the execution.
I was able to overcome this issue replacing the HashMaps by the use setAttributes for each solution.
This was the approach originally used by the MOEA/DD on jMetal 4.5 version.

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/moead/MOEADBuilder.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/moead/MOEADBuilder.java
@@ -15,7 +15,7 @@ import org.uma.jmetal.util.AlgorithmBuilder;
  * @version 1.0
  */
 public class MOEADBuilder implements AlgorithmBuilder<AbstractMOEAD<DoubleSolution>> {
-  public enum Variant {MOEAD, ConstraintMOEAD, MOEADDRA, MOEADSTM} ;
+  public enum Variant {MOEAD, ConstraintMOEAD, MOEADDRA, MOEADSTM, MOEADD} ;
 
   protected Problem<DoubleSolution> problem ;
 
@@ -186,6 +186,10 @@ public class MOEADBuilder implements AlgorithmBuilder<AbstractMOEAD<DoubleSoluti
     } else if (moeadVariant.equals(Variant.MOEADSTM)) {
         algorithm =  new MOEADSTM(problem, populationSize, resultPopulationSize, maxEvaluations, mutation,
                 crossover, functionType, dataDirectory, neighborhoodSelectionProbability,
+                maximumNumberOfReplacedSolutions, neighborSize);
+    } else if (moeadVariant.equals(Variant.MOEADD)) {
+        algorithm = new MOEADD(problem, populationSize, resultPopulationSize, maxEvaluations, crossover, mutation,
+                functionType, dataDirectory, neighborhoodSelectionProbability,
                 maximumNumberOfReplacedSolutions, neighborSize);
     }
 
--- a/jmetal-exec/src/main/java/org/uma/jmetal/runner/multiobjective/MOEADDRunner.java
+++ b/jmetal-exec/src/main/java/org/uma/jmetal/runner/multiobjective/MOEADDRunner.java
@@ -1,0 +1,89 @@
+package org.uma.jmetal.runner.multiobjective;
+
+import org.uma.jmetal.algorithm.Algorithm;
+import org.uma.jmetal.algorithm.multiobjective.moead.AbstractMOEAD;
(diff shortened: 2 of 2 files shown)


jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/moead/MOEADD.java
```diff
@@ -232,7 +221,7 @@ public void updateArchive(S indiv) {
 
     // find the location of 'indiv'
     setLocation(indiv, idealPoint, nadirPoint);
-    int location = regionSolution.getOrDefault(indiv, 0);
+        int location = (int) indiv.getAttribute("region");
```
matthieu-vergne: Same thing everywhere, so I won't repeat it anymore. Please check all your indentation.

Meanwhile, this line is among the relevant changes of the pull request: you get the attributes from the individual itself rather than some global containers. The actual issue with maps is that they store the information *ad vitam eternam*. With your change, the individual stores the information by itself so it disappears with it.

The first question is whether or not we want to store the information indefinitely. If we assume that the same solutions might be generated really often, then it is interesting because we save computation time by not regenerating the same values. But here it seems that it does not happen that often. So you were right by choosing a way which makes the information disappear with the individual. However, you did not need all these changes for that: just replacing `HashMap` by `WeakHashMap` allows to let the garbage collector get rid of individuals not used anymore. So if I am correct, just replacing the three maps by weak maps should be enough to fix the problem.

I did not try it, but I would recommend to try with this small change first to see if it fixes the issue. If it does not, then there is something else going wrong, and this pull request might be worth to merge (once the indentation and generics have been fixed).

jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/moead/MOEADD.java
```diff
@@ -250,7 +239,8 @@ public void updateArchive(S indiv) {
             lastFront.add((S) population.get(i));
           }
         }
-        if (rankSolution.getOrDefault(indiv, 0) == (numRanks - 1)) {
+                if (((int) indiv.getAttribute(ranking.getAttributeIdentifier())) == (numRanks - 1)) {
+//        if (rankSolution.getOrDefault(indiv, 0) == (numRanks - 1)) {
```
matthieu-vergne: Please, don't comment old code. If it is wrong, delete it. If it is not, keep it. If you don't know, test it. The whole point of versioning the code is to retrieve it if necessary through the commit history, so don't let commented code.

jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/moead/MOEADD.java
```diff
@@ -1320,9 +1312,9 @@ public void setLocation(S indiv, double[] z_, double[] nz_) {
       }
     }
     //indiv.setRegion(minIdx);
-    regionSolution.put(indiv, minIdx);
+        indiv.setAttribute("region", minIdx);
     //indiv.Set_associateDist(minDist);
-    associateDistSolution.put(indiv, minDist);
+//        indiv.setAttribute(ATTRIBUTES.DIST, minDist);
```
matthieu-vergne: Why is this one commented?

jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/moead/MOEADD.java
```diff
@@ -1478,22 +1470,19 @@ public int countTest() {
...
   }
 
   public void replace(int position, S solution) {
     if (position > this.population.size()) {
       population.add(solution);
     } else {
       S toRemove = population.get(position);
-      rankSolution.remove(toRemove);
-      regionSolution.remove(toRemove);
-      associateDistSolution.remove(toRemove);
```
matthieu-vergne: Explicit removal becomes useless with weak maps, so these lines can be removed as well, although they can also stay there. I don't know whether explicit removal increases performance.

matthieu-vergne:
Did you try with weak maps? Just taking the version before your change, replace the maps you remove by weak maps and running it, does it fix the problem?

I agree that attributes are another solution, but on the long term view, I think that this form of attributes will disappear.

matthieu-vergne:
Of course, if weak maps do not fix the problem, then I agree with your solution, which is rather equivalent in principle.

fritsche:
I have submitted a pull request https://github.com/jMetal/jMetal/pull/236 implementing a version using WeakHashMaps. Either the pull request #233 or #236 solves the OutOfMemoryIssue

PR #236, opened 2017-11-17
MOEADD using WeakHashMaps

fritsche:
As discussed by @matthieu-vergne at https://github.com/jMetal/jMetal/pull/233
Another possible way to solve the OutOfMemory from MOEADD is the use of WeakHashMaps.
This Pull Request implements this solution proposed by @matthieu-vergne.

* It replaces the use of solution attributes (previously stored in HashMaps) by the use of WeakHashMaps.
* Includes MOEAD Weights for many objective problems
* Includes an example of execution of MOEADD for many objective problems


matthieu-vergne:
Thank you for your effort, but I don't understand why it is about replacing attributes by weak maps, since master is supposed to use normal maps, so the changes should have been 3 lines. Here it is way longer, so it is hard to review and will take some time. So I will look at it later.

Additionally, I saw that there is also some code formatting. PLEASE don't do that, never, it just adds useless changes to review. If at some point there is the need to format a file, it can be done in a separate, dedicated pull request. Don't forget that the more complex your pull request is, the less chance you have people look at it. Everyone is busy, give them only the relevant stuff.

matthieu-vergne:
OK, now I understand: @ajnebro already integrated your changes before I do my review. Well then, it is already integrated, so already fixed. We will change it again later if needed. Sorry for the noise.

fritsche:
Since it was already fixed, I think we can close this one.

PR #249, opened 2018-03-07
Now WASFGA and GWASFGA check if weight vectors have a valid format.

rsain:
I am applying for this pull request because GWASFGA and WASFGA must check if the weight vectors are consistent regarding the problem being solved.

Weight vectors are valid (or consistent) if they have as many components as objectives the problem being study has. I have added this check and now the algorithms throw a JMetalException whether weight vectors are invalid.


matthieu-vergne:
Please focus your pull requests. You say you apply this because of some checks, but you actually do many unrelated things too. Such pull requests are a pain to review (and thus prone to never be accepted). So please favor creating multiple pull requests, each focused on a single feature, rather than few big ones.

rsain:
Hi @matthieu-vergne. Thank you for your feedback. I totally agree with you and I will do that next time.

In addition to what I commented in the description of the pull request, I also (1) applied some automated refactoring proposed by the IDE and (2) modified and renamed the class WeighVector. Although all is related, there are a lot of changes and it makes hard to review the code.

Thanks for the advise. I will make easier the job next time ;)

Regards,
Rubén.

matthieu-vergne:
Thanks.

rsain:
Hi @matthieu-vergne. I know we are all busy, but I was wondering if this pull request is going to be accepted or you want me to create multiple pull requests instead.

We are working in a new version and more pull requests are coming ;)

Cheers.

matthieu-vergne:
Unfortunately I have no authority on that level. I am but a mere contributor here. I don't know whether @Juanjdurillo has any rights to merge pull requests, but the owner of the repo is @ajnebro, so you may need to wait for him to consider this pull request.

Surely enough, splitting it would increase your chances, although I don't know how much. So if you have time to do so, I would recommend it.

ajnebro:
All the pull requests of @rsain are very welcome :)

PR #283, opened 2018-08-20
Update README.md

matthieu-vergne:
Split status section + add how to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,30 @@
 # jMetal 5 Development Site
 
-### Status
+## Purpose
+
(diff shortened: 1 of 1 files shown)


ajnebro:
That information is included in the documentation site: https://github.com/jMetal/jMetalDocumentation/blob/master/installation.md

Sure it should be updated and improved.

matthieu-vergne:
The local readme should be safe contained regarding the project. Usage description of this project should not be in another project. Even if it is just to provide a link to the documentation. Like:
> To know how to use this project, please refer to the [installation guide](https://github.com/jMetal/jMetalDocumentation/blob/master/installation.md).

PR #290, opened 2018-09-25
Dynamic Problem

jacksonpradolima:
Hello,
I created a "core" for dynamic problems where we needs a solutions with variable size.

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/problem/impl/AbstractFlexibleVariableProblem.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/problem/impl/AbstractFlexibleVariableProblem.java
@@ -1,0 +1,76 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2018 Jackson Antonio do Prado Lima <jacksonpradolima at gmail.com>.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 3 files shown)


jmetal-core/src/main/java/org/uma/jmetal/solution/IFlexibleVariableSolution.java
```diff
@@ -0,0 +1,119 @@
...
+ * Interface representing a solution with dynamic size variables. This interface
+ * is used when it is necessary to manipulate the number of variables of a
+ * solution. For example, a solution may have 5 variables and other solution 2
+ * variables, after add, remove or set some variable.
+ *
+ * @author Jackson Antonio do Prado Lima <jacksonpradolima at gmail.com>
+ * @version 1.0
+ * @param <T>
+ */
+public interface IFlexibleVariableSolution<T> extends Solution<T> {
```
matthieu-vergne: Please use the local conventions. We don't prefix interfaces with `I` here.

jmetal-core/src/main/java/org/uma/jmetal/problem/impl/AbstractFlexibleVariableProblem.java
```diff
@@ -0,0 +1,76 @@
...
+import org.uma.jmetal.solution.IFlexibleVariableSolution;
+
+/**
+ *
+ * @author Jackson Antonio do Prado Lima <jacksonpradolima at gmail.com>
+ * @version 1.0
+ * @param <S> Solution type
+ */
+@SuppressWarnings("serial")
+public abstract class AbstractFlexibleVariableProblem<S extends IFlexibleVariableSolution<?>> extends AbstractGenericProblem<S> {
```
matthieu-vergne: This implementation does not do what we expect from it. `AbstractGenericProblem` does provide setters, but they are not bounded by the min/max we define here. The limits are just totally ineffective.

Like you had to redo the implementation of `Solution`, you need to redo the one for `Problem`. What you can do, however, is to make the existing implementations extend your own, which would drastically reduce code redundancy.

PR #334, opened 2019-02-20
Restrict experiments to fulfill the requirements of ExperimentAlgorithm.

matthieu-vergne:
# Context

`ExperimentAlgorithm` accepts any type of `Algorithm`, in the sense that it does not constrain its type of `Result`.

# Problem

This openness is not genuine, since it does cast the result into a list of solutions:
```java
new SolutionListOutput((List<S>) population) // <- Only List<S> is accepted
        .setSeparator("\t")
        .setVarFileOutputContext(new DefaultFileOutputContext(varFile))
        .setFunFileOutputContext(new DefaultFileOutputContext(funFile))
```

# Changes

I constrain the generics to ask for results in the form a of a list of solution. It allows to remove the cast, since it is now guaranteed to have the right type.

However, the indirect consequence is to break many experiment-related classes which did not assume such restriction. I spread this constraint all over the concerned classes to fix that, aligning it with the type of solution provided when relevant. Thus, most of the changes are about adding constraints to experiment.

As a consequence, all the jMetal experiments (`Experiment` class) make the assumption of dealing with algorithms generating a list of solution. Indeed, this root class directly builds on `ExperimentAlgorithm`, so it seems hard to assume that it can do without this constraint.

# Question

@ajnebro can you confirm that you only used these experiments with lists of solutions? If it is the case, then it should be OK to merge this change. Otherwise, could you please share an example of experiment on an algorithm providing a different type of result? This way I could see how to reduce the impact in a pragmatical way.

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/Experiment.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/util/experiment/Experiment.java
@@ -13,7 +13,7 @@ import java.util.List;
  *
  * Created by Antonio J. Nebro on 17/07/14.
  */
-public class Experiment<S extends Solution<?>, Result> {
+public class Experiment<S extends Solution<?>, Result extends List<S>> {
 	private String experimentName;
 	private List<ExperimentAlgorithm<S, Result>> algorithmList;
 	private List<ExperimentProblem<S>> problemList;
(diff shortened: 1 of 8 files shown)


ajnebro:
The experiments are intended to compare multi-objective algorithms, and they always return list of solutions.

matthieu-vergne:
So I guess it is correct to constrain the results to be lists of solutions, then.

If it is the case (and if you like them) you can merge this commit.

PR #338, opened 2019-02-25
Fix warnings

matthieu-vergne:
Here is another resolution of warnings. Some are trivial stuff, but others not.

For closing the resources, I used an explicit close or a try-resource pattern depending of what I judged more practical. The second solution implies a reindentation, which is why you can see big blocks of changes. If you hide whitespaces, you can see that it is just about closing the resources.

For the unchecked types, I added explicit variables to suppress warnings on them (to not suppress it on the whole method).

For unused stuff, I have **removed several variables, fields, and methods**. Some removals made public methods useless. I did not remove them to not break anything, but I added deprecated tags. Please look at these changes to see if the removal are properly motivated.

All the other commits are about trivial stuff.


matthieu-vergne:
\\{^o^}/ Warning hunt is over!

PR #340, opened 2019-03-24
Implementations of First Ascent Hill Climb, Steepest Ascent Hill Climb and Simulated Annealing

jolmedilla:
...plus operators and utility classes needed for them.
These implementations were done as part of my PhD dissertation work in University College Dublin directed under Mel Ó Cinnéide. This is the result of porting my original work back to jMetal.

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/firstascenthillclimb/FirstAscentHillClimb.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/firstascenthillclimb/FirstAscentHillClimb.java
@@ -1,0 +1,91 @@
+/**
+ * Copyright 2016 Juan Olmedilla Arregui
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.uma.jmetal.algorithm.singleobjective.firstascenthillclimb;
+
+import java.util.Comparator;
+
+import org.uma.jmetal.algorithm.impl.AbstractHillClimb;
+import org.uma.jmetal.operator.LocalSearchOperator;
+import org.uma.jmetal.operator.MutationOperator;
+import org.uma.jmetal.operator.impl.localsearch.BasicLocalSearch;
+import org.uma.jmetal.operator.impl.localsearch.LocalSearchWithMutationReverse;
+import org.uma.jmetal.problem.Problem;
+import org.uma.jmetal.solution.Solution;
+import org.uma.jmetal.util.comparator.FitnessComparator;
+
+/**
+ * First Ascent Hill Climb: Single objective basic local search Time bound
+ * search.
+ * 
+ * @author Juan Olmedilla Arregui
+ *
+ */
+public class FirstAscentHillClimb<S extends Solution<?>>
+        extends AbstractHillClimb<S> {
+
+    /**
+     * 
+     */
+    private static final long serialVersionUID = 1L;
+
+    public FirstAscentHillClimb(int mutations,
+            MutationOperator<S> mutationOperator,
+            MutationOperator<S> rollbackOperator, Problem<S> problem,
+            int durationInMinutes) {
+        super(mutations, mutationOperator, rollbackOperator, problem,
+                durationInMinutes);
+    }
+
+    /**
+     * Constructor with no reverse operator. This is meant for all other cases
+     * in which the Solutions do not share important information.
+     * 
+     * @param mutations number of maximum mutations we allow
+     * @param mutationOperator the mutation operator
+     * @param problem original problem
+     * @param durationInMinutes duration in minutes
+     */
+    public FirstAscentHillClimb(int mutations,
+            MutationOperator<S> mutationOperator, Problem<S> problem,
+            int durationInMinutes) {
+        this(mutations, mutationOperator, null, problem, durationInMinutes);
+    }
+
+    @Override
+    public String getDescription() {
+        return "First Ascent Hill Climb";
+    }
+
+    @Override
+    public String getName() {
+        return "HCF";
+    }
+
+    @Override
+    protected LocalSearchOperator<S> getNewLocalSearch() {
+        LocalSearchOperator<S> search;
+        Comparator<S> comparator = new FitnessComparator<S>();
+        if (rollbackOperator != null)
+            search = new LocalSearchWithMutationReverse<S>(1,
+                    mutationOperator, rollbackOperator, comparator,
+                    problem);
+        else
+            search = new BasicLocalSearch<S>(1, mutationOperator,
+                    comparator, problem);
+        return search;
+    }
+
+}
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/simulatedannealing/SimulatedAnnealing.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/simulatedannealing/SimulatedAnnealing.java
@@ -1,0 +1,221 @@
+/**
+ * Copyright 2016 Juan Olmedilla Arregui
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.uma.jmetal.algorithm.singleobjective.simulatedannealing;
+
+import java.util.Comparator;
+
+import org.uma.jmetal.algorithm.impl.AbstractSingleSolutionSequentialAlgorithm;
+import org.uma.jmetal.operator.LocalSearchOperator;
+import org.uma.jmetal.operator.MutationOperator;
+import org.uma.jmetal.operator.impl.localsearch.BasicLocalSearch;
+import org.uma.jmetal.operator.impl.localsearch.LocalSearchWithMutationReverse;
+import org.uma.jmetal.problem.Problem;
+import org.uma.jmetal.solution.Solution;
+import org.uma.jmetal.util.pseudorandom.JMetalRandom;
+import org.uma.jmetal.util.solutionattribute.impl.Fitness;
+
+/**
+ * Single Objective Simulated Annealing
+ * 
+ * @author Juan Olmedilla Arregui
+ *
+ */
+public class SimulatedAnnealing<S extends Solution<?>>
+        extends AbstractSingleSolutionSequentialAlgorithm<S> {
+
+    /**
+     * 
+     */
+    private static final long serialVersionUID = 1L;
+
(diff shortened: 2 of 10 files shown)


matthieu-vergne:
If @ajnebro already checked everything and you pushed it for him to integrate it, just ignore my comment.

Could you please make more atomic pull requests, so we can review each algorithm separately? By making huge pull requests, we must check all at once, which just makes it less likely that we do it at all. At least, I won't check 3 algorithms at once.

My recommendation would be:
1. Split your commit:
  - 1 commit per utility
  - 1 commit per algorithm
2. Create dedicated PRs:
  - create a branch from master
  - cherry-pick the relevant commits for a specific algorithm (not a problem if a commit appears in several branches)
  - create a pull request for it
  - redo until you have all your algorithms

Meanwhile, it seems there is some compilation issues with Travis. You may give it a look first.

jolmedilla:
Hi,

I can break the commits into every algorithm but the errors are related with some abstract classes   used by them all. Once the abstract classes are in, the actual algorithms will be very small.
As for the compilation errors, I’m afraid that they have more to do with the way the compiler works in Travis. Perhaps you can help me on this:

Compiler complains that a class uses interfaces and classes that do not exist, but that is not so, they are in the same project but they are compiled later due to their alphabetical order. This is only happening in the Travis job. The problem is that if I re-arrange names and classes  in different packages I will break the logic of placement in packages according to meaning.
Can you take a look and help on this?
Thanks.

Juan J. Olmedilla

matthieu-vergne:
First, your commit is based on an old master, so please rebase your work on the current master branch to resolve 169 warnings. I just made a new PR to fix more of them, but it may take some time to be integrated, so let's not rely on it.

Then I don't have any compilation issues on my side, so I would ignore Travis for the time being and look at it if it reproduces on the new branches.

Regarding the fact that all your algorithms are linked through various abstractions, it is an issue itself, because it means that if I want to understand the gist of what FirstAscentHillClimb is doing, reading FirstAscentHillClimb.java is not enough, because this file only describes a name, a description, and a local search, and I have no clue about how they are used without digging in the class hierarchy. Hill climbing is supposed to be a simple algorithm. Yet here, to understand what FirstAscentHillClimb is doing, I need to understand the leaf class + 3 abstract classes it extends. Especially, you override several times `initProgress()` and `updateProgress()`, override several times `run()` and then switch to an `execute()`, etc. So you just cannot understand what your algorithm is doing without looking at the details of each class of the hierarchy. That is just overkilling!

As a general remark, please don't use abstract classes to factor the code unless it is the most relevant choice (i.e. almost never). It introduces dependencies everywhere (protected fields just make it worse) which makes it harder to understand, harder to maintain, and less flexible for fixes and optimizations. Don't forget that in Java we can only extend 1 class! So we cannot compose abstract classes having different interesting features. Factoring should be done first through dedicated (non abstract) classes which focus on the specific feature you want to factor, and then inserted in the algorithm, not extend them. JMetal is full of that kind of mistakes, but that is not the right way to go if we seek maintainability. You may read [this post](https://dzone.com/articles/when-to-use-java-abstract-classes) for another view on that matter.

If @ajnebro  want to integrate it, he will do it. In my case, I won't review this code before I see a PR which provides a single class for the algorithm (not a hierarchy with abstract classes) + optionally some utilitary stuff relevant for it. Factoring can come later if semantically motivated.

jolmedilla:
Hi Mathieu,

Thanks for your prompt response.
As for your remarks, I have a question:

There are two algorithms: First Ascent Hill Climb and Steepest Hill Climb, both are Hill Climb type algorithms. Following your advise I will not have an abstract class from which both inherit common behaviour but two completely unrelated classes, right? Then I will have no choice other than copying and pasting common behaviour in both classes, which I believe is even a worse practice than using an abstract class. I know you will tell me that you can extract the common behaviour to some decorator or some class that can be used via composition, for instance a Strategy pattern, but that would be a bit weird given the fact that in jMetal, an in general in optimisation algorithms, the Algorithm is the strategy.
Believe I have given it a very thorough thought and I am using abstract classes in this cases because it makes sense. 
By the way, my research is about automation of Object Oriented Refactorings and the heuristic we use in jMetal is Object Oriented Quality metrics.
Thanks,

Juan J. Olmedilla

matthieu-vergne:
They are not necessarily unrelated. You could fairly imagine to have a generic HillClimbing class which implements the generic stuff, and to build on it to implement several variants. You can do it through abstract classes or by using the template method strategy. This is the pattern that jMetal may use heavily instead of abstract classes everywhere. For HillClimbing, both are good, choosing is more a matter of preference.

My remark is about the multiplication of abstraction levels to deal with highly specific stuff (AbstractSingleSolutionSequentialAlgorithm? AbstractMeasurableTimeLimited? wtf!?). This is painful to understand and to maintain. We are dealing with humans here, not machines. It must remain understandable. These classes smells bad because they speak about features that should not be entangled to specific algorithms, but rather be features ready to be used in any algorithm that want to integrate it. Abstract classes are not good for such cases.

Since you work in automation of refactorings, maybe you will get my point: refactoring is not about syntax, but semantics. It is not because two codes are identical at a given time that they must be factored into the same method or class. This is because they *must* be the same that it makes sense to create a single method on which both rely. If changing one implies to change the other, they are semantically linked and thus should build on the same pieces. If changing one can be done without *requiring* to change the other, then it makes no sense to factor them, even if at that point in time they appear to be exactly the same.

Copy-paste is not a bad thing, since later it may evolve differently. At best, noticing two similar codes should *suggest* to factor them. This is by analysing the involved *concepts* that you confirm whether or not it makes sense. Too often I see people factoring stuff that should not, because it appeared at that time that they had the same code. By reflecting more on the *purpose* of each code, they would have noticed that they shouldn't, and that the code could have been slightly different to properly consider the specificities of each case. The codes had to be different if we wanted to optimize them, but once they were factored, these optimizations were not possible anymore because optimizing for one would have worsen the other.

The point is that deciding whether they are linked in that way is all about semantics, not syntax. People who rely on automation to factor stuff should not do it blindly, otherwise we get PRs like these ones [[1]](https://github.com/jMetal/jMetal/pull/271) [[2]](https://github.com/jMetal/jMetal/pull/270) [[3]](https://github.com/jMetal/jMetal/pull/269). Maybe you know them? It seems like the work of some of your colleagues.

My goal when programming is to ensure not only that it works, but also that the people who will pass after me will be able to maintain what I did. Blind automation goes against that goal, as well as stacks of abstract classes. I love abstraction, I made plenty of these mistakes, and I learned it the hard way. Then I reviewed these PRs, and it was painful again. Now I want to save myself from this kind of useless work. You want to automate? Great! Now be sure that it reduces my work, not increase it.

I am also a researcher, although not in your domain, but I know fairly well that being a researcher in a domain does not make you automatically flawless on the matter you are researching. Automation is nice, but did you consider the understandability and maintainability of the code you produce? I don't say you did not, but the previously cited merge requests make me think that the author of these commits did not consider such criteria. From my opinion, these criteria are fundamentals to make good refactoring. As long as humans are the ones to maintain the code. If one day the code is fixed and evolved by machines, then it won't be a relevant criteria anymore, and at that time refactoring might be less interesting because it is all a matter of optimization and tests (or proofs).

Now, what I say is that your code can be completely refactored to have something better. Maybe you don't care, and that would be fine because everyone have their own values. Maybe you agree, and then you can do it. Maybe you don't know but you are interested, in which case I don't mind working on it to show you how I would do it, but that will take some time. In any case, I have no authority here. I am only a perfectionist who try to share what he learned.

jolmedilla:
Well, we have radically different views about Copy-paste. I tend to follow Martin Fowler in this. 
Anyway, sorry to have bothered you.
Best regards,

Juan

matthieu-vergne:
Sorry for the aggressivity of my post. Your code is far from professional level (e.g. no documentation nor automated tests), and reading "my research is about automation of Object Oriented Refactorings" looked to me like I should take your code as some kind of Graal. In addition to a bad experience with automated refactoring PRs.

Could you cite a reference regarding Martin Fowler? When I search about copy-paste with his name I find that:
https://martinfowler.com/bliki/Seedwork.html

jolmedilla:
Hi Mathieu,

You are absolutely right about my code lacking many things and that’s the kind of comments I was expecting, in fact, I was surprised you did not bring up the lack of tests firstly and that you focused in something that is, in my opinion, debatable. Let me summarise my points here, all in a spirit of collaboration:

- I can agree about some of my abstract classes being subject to refactoring, especially the ones you mentioned in your last e-mail, the ones that are more general, in particular AbstractSingleSolutionSequentialAlgorithm and AbstractMeasurableTimeLimited. I can work on that.
- As for the two that are part of the family of Hill Climb algorithm I still disagree. And I am of the same line of thought as yourself in the sense that is a semantic drive what leads the whole thing. However, for implementing a Factory Method pattern you have to have an abstract class given the fact that the abstract method is declared in a parent class that invokes it in a non abstract method. That is the whole point of placing what is really common to all Hill Climb algorithms in such abstract class, sharing a common strategy.
- Regarding Copy-paste, I cannot tell you the exact page in Martin Fowler’s book (https://www.amazon.com/Refactoring-Improving-Existing-Addison-Wesley-Signature/dp/0134757599/ref=sr_1_1?keywords=refactoring&qid=1553503307&s=gateway&sr=8-1 <https://www.amazon.com/Refactoring-Improving-Existing-Addison-Wesley-Signature/dp/0134757599/ref=sr_1_1?keywords=refactoring&qid=1553503307&s=gateway&sr=8-1>) but it is a well known bad smell mentioned there somewhere and if not, check Riel’s book (https://www.amazon.com/Object-Oriented-Design-Heuristics-paperback-Arthur/dp/0321774965 <https://www.amazon.com/Object-Oriented-Design-Heuristics-paperback-Arthur/dp/0321774965>). The idea that you can copy and paste and that both chunks of code can evolve separately is precisely what leads to very unmaintainable code, if they are really different you do not need to copy them, if they are not, you can either use some strategy pattern, decorator or some pattern based on composition that will lead you to being able to really share the common code and still avoid regressions.
- As for the documentation I still remember Fowler’s refactoring for that bad smell: convert comment into method (cannot tell the exact page). The classes and methods must be readable enough, if not, add more methods, rearrange things. My code is far from perfectly understandable and of course, I am open to any suggestions. When I first uploaded it, I did not expect to go without any request for change on your part.
- In general, about what you mention of having all the code of an algorithm in one single class and never ever using abstract classes, well, I’m afraid I still disagree in that. That still leads to a procedural way of coding in which you have massive classes. Yes, OO code is more complex, yes you have to jump classes to understand the whole thing, but the important thing is that you understand the class in isolation, let me explain with an example: if you have a class called FirstAscentHillClimb that extends AbstractHillClimb you must understand the latter by itself, but only the behaviour that is specific of it as per the Single Responsibility Principle, the other classes that it extends or it uses (interfaces in this last case) must have names that are sufficiently self-explanatory of what their responsibility is.

Sorry If I cannot be more precise when citing my sources, it’s being a long time since I was in research. In my day I breakfast, lunched and dined the bloody OO thing and I have the automatics of it but I cannot go to the exact page. I just wanted to contribute back part of what I had done in my day. I do not expect to go without changes but I have very limited time to dedicate to this. I did not want to sound like what I did was the Graal but telling me that I simply could not use abstract classes at all and that copy-and-paste was acceptable in some cases sounded to me like we were too far apart to make this workable. If this is the case, I have no trouble in stepping aside and leaving the whole thing in the most amicable way.
Thank you and regards,

Juan Olmedilla

ajnebro:
Contributions of researchers to the jMetal project are always welcome, and that's something to be thankful for.  Sometimes, that contributed code does not fit into a minimum of quality but, whenever it does not interfere with existing classes, I usually acept it.

ajnebro:
I still have to check this pull request to take a decision

matthieu-vergne:
> in fact, I was surprised you did not bring up the lack of tests firstly and that you focused in something that is, in my opinion, debatable.

Surely design arguments are debatable. The point is that they are the only kind of comments that seem to get through, here. Since I joined this project, I don't remember any PR with automated tests beside mines. I did ask for tests when reviewing the PRs at the start, but since it was systematically ignored or disregarded, I just stopped asking for them.

> As for the documentation I still remember Fowler’s refactoring for that bad smell: convert comment into method

Comments and documentation are different things. I speak about Javadoc, not comments. I fully agree the latter should not be there, unless they provide a clear explanation of some subtle optimization made and why, which should be exceptional. Javadoc, however, should be systematic on "directly accessed" items (interface declarations, interface methods, class declarations, constructors). The name tells what it is/does, while the documentation can give additional info like why it is there, examples of use, and so on.

For the other points, thank you for the pointers. I will check on them and see if I can get something interesting out of them.

PR #378, opened 2019-11-25
Set JDK to openjdk8 in Travis, as oracle no longer works

mbuzdalov:
As long as I know, Oracle JDK 8 is no longer available at Travis, so continuous integration basically does not work.

One can change the JDK to OpenJDK 8 in order to fix that. This is what this PR is about.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 install: mvn compile
 jdk:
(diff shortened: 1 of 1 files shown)


ajnebro:
Hi Maxim.
  Thanks for the pull request.

Antonio

PR #379, opened 2019-11-25
Implement faster dominance ranking

mbuzdalov:
This is an experimental attempt to use an external library, `com.github.mbuzdalov:non-dominated-sorting-implementations` version 0.2.1, to perform dominance ranking in jMetal.

An adapter, called `ExperimentalFastDominanceRanking`, is written to implement `Ranking` similar to `o.u.j.u.s.i.DominanceRanking`. This adapter, unlike `DominanceRanking`, directly uses objective values and `OverallConstraintViolation` as an additional objective (when needed). This is for performance; however, various possible semi-correct hacks, such as epsilon-dominance, are not supported by this class.

Inside it delegates most of the work to a `NonDominatedSorting` from the library above. It is created by a `NonDominatedSortingFactory`, which determines which particular algorithm is to be used to do sorting. As of right now, one particular factory is used, although it is possible to determine it on the fly. 

To make sure both rankings understand basic things similarly, a test suite for `DominanceRanking` is copied over the new implementation. However, the tests there are too basic to capture anything serious, including constraint violation, so some work probably needs to be done here.


matthieu-vergne:
Hi, Thank you for your effort.

Do you have some benchmark to show the difference of performance?

mbuzdalov:
Sorry for not doing it using JMH, but the results shall be pretty close.
Notation: "old" for `DominanceRanking`, "new" for submitted `ExperimentalFastDominanceRanking`. 

1) Test with points uniformly sampled from a hypercube (aka the "cloud dataset").

N, algo | D = 2 | D = 3 | D = 5 | D = 8
----|----|----|----|----
100, old | 0.000168 | 0.000157 | 0.00016 | 0.00017
100, new | 0.000008 | 0.000015 | 0.000022 | 0.00004
1000, old | 0.0275 | 0.0185 | 0.0153 | 0.0175
1000, new | 0.0002 | 0.00056 | 0.0011 | 0.0013

2) Same with objective 0 set in such a way that sum of all objectives is 0 (a poor man's planar test).

N, algo | D = 2 | D = 3 | D = 5 | D = 8
----|----|----|----|----
100, old | 0.000118 | 0.000123 | 0.000143 | 0.000175
100, new | 0.000007 | 0.000016 | 0.000029 | 0.000031
1000, old | 0.0117 | 0.0117 | 0.0138 | 0.017
1000, new | 0.0001 | 0.0004 | 0.001 | 0.0012

matthieu-vergne:
Could you share your benchmark code?

matthieu-vergne:
You can write it in the `jmetal-core/src/test/resources/` folder.

mbuzdalov:
Will `jmetal-core/src/test/java/org/uma/jmetal/util/solutionattribute/impl` folder suffice? It is an ordinary public-static-void-main-type class. I believe it would not be easily run from `resources`.

mbuzdalov:
I have added the benchmarking code next to the tests (this does not break how JUnit works).

In the meantime, I found out that my understanding of constraint violation comparators was wrong, and I fixed it.

matthieu-vergne:
> Will `jmetal-core/src/test/java/org/uma/jmetal/util/solutionattribute/impl` folder suffice? It is an ordinary public-static-void-main-type class. I believe it would not be easily run from `resources`.

Even better. {^_^}b
I thought it was some scripts, so I suggested the resources.
If it is some Java code it surely helps to understand each other.

> In the meantime, I found out that my understanding of constraint violation comparators was wrong, and I fixed it. By the way, unit tests don't check how `DominanceRanking` deals with constraint violations.

jMetal is actually quite poor in terms of testing.
It is highly variable, some classes are well tested and others not at all.
Any participation on that matter is greatly appreciated.
If you spot missing tests, don't hesitate to add them.

I plan to go through the code of the rankings to compare your code with ours.
But since I am not that familiar with the original one, I will need to dig in the details of each class.
The process might go through adding more tests too.
I plan to do that when I find some time this week.

ajnebro:
Adding a new dominance ranking will require to refactor some components to allow choosing between using the default one, the one proposed by Maxim, and other that some colleagues have developed.

matthieu-vergne:
This is something we can do in a separate issue.
Fundamentally, it is not hard to add a constructor with an additional parameter.
Nothing should break with this kind of change.

ajnebro:
I have adapted Maxim's code to include it in the [jmetal6 branch](https://github.com/jMetal/jMetal/blob/jmetal6/jmetal-auto/src/main/java/org/uma/jmetal/auto/util/ranking/impl/ExperimentalFastNonDominanceRanking.java). To test it, I have used the [NSGA-II implementation](https://github.com/jMetal/jMetal/blob/jmetal6/jmetal-auto/src/main/java/org/uma/jmetal/auto/algorithm/nsgaii/NSGAII.java) included in the jmetal-auto package. 

I have measured computing times using NSGA-II with standard settings (population size and offspring population size = 100, number of evaluations = 25000) to solve the ZDT1 problem. The times obtained are (roughly):

- Fast NDS (NSGA-II): 890 ms
- Experimental Fast NDS: 605 ms

When changing the value of the offspring population size to 1 (to configure a steady-state version), the times are:

- Fast NDS (NSGA-II): 3650 ms
- Experimental Fast NDS: 2400 ms

So the time reductions with these experiment settings are around 33%, which is quite impressive. 

However, I have noticed, by fixing the random number generator seed, that using the two NDS algorithms produce different fronts, which at a glance seems counterintuitive. Anyway, I have to adapt and check yet the unit tests.

mbuzdalov:
> However, I have noticed, by fixing the random number generator seed, that using the two NDS algorithms produce different fronts, which at a glance seems counterintuitive. Anyway, I have to adapt and check yet the unit tests.

One possible cause is that the order of solutions in subfronts may matter, especially if there are some solutions with identical fitness vector, or some solutions with identical crowding distance. I am not sure if this can be easily unified from my side.

Or maybe my understanding of constraint violation handling is still wrong. Anyway, the more unit tests, the better.

What I can do now is to implement a way to compare the existing algorithms on a large number of random test cases; if they appear to be fundamentally different, it is something to analyze.

> When changing the value of the offspring population size to 1 (to configure a steady-state version), the times are

This is what I aim at doing on the next phase by introducing proper data structures, which enable the approach of Ilya to be integrated. My aim is to get the steady-state times back to ~600-800 milliseconds.

mbuzdalov:
A quick update:

1) The two implementations of `Ranking` behave identically regarding the ranks of the points, when there is no constraint violations.

2) They do differ when there exists constraint violation, and I don't yet know why.

3) However, the difference in fronts which @ajnebro observes may be simply because:

```
    public class RankingComparator<S extends Solution<?>> implements Comparator<S>, Serializable {
      private Ranking<S> ranking = new DominanceRanking<S>() ;
```

the important comparators still use the old ranking, and the identity of the properties is simply classname-based. So they basically see no ranks which have been computed by my code.

mbuzdalov:
I have found and fixed problems with accounting for overall constraint violation in my implementation. The corresponding test is also introduced, as well as the torture testing suite to compare the implementations.

The problem with having `DominanceRanking` as the only property across the codebase still persists. Maybe we need to:
* have a singleton`DominanceRanking` which is a `SolutionAttribute`;
* have an interface `DominanceRankingAssignment` which is for algorithms that assign ranking;
* all the implementations, both the old `DominanceRanking` and my new code, implement the latter and set the `DominanceRanking` property?

What is more, we have the `Ranking` interface which is mostly for `DominanceRanking` in most of the algorithms, except for a few. Maybe the idea of an abstract `Ranking` with no underlying concept is not so healthy...

ajnebro:
I'm not very happy with the approach I adopted in jMetal 5 to organize entities such as rankings and density estimators as part of the solution attributed package.

In the jmetal6 branch I'm exploring another organization where there is a jmetal.util.ranking package containing the `Ranking` interface: 

```java 
public interface Ranking<S> extends Attribute {
  Ranking<S> computeRanking(List<S> solutionList) ;
  List<S> getSubFront(int rank) ;
  int getNumberOfSubFronts() ;
}
```

and a number of implementations; currently: the NSGAII's NDS, SPEA2's strength, and now Maxim's NDS.

mbuzdalov:
Again, if the new `Attribute` is much the same as the old `SolutionAttribute` with the class-name identity -- which, as per `private String attributeId = getClass().getName();`, is still true in jmetal6 -- both implementations will write to different property slots of the solution. If NSGA-II's comparators will pick one of them to rely on, the output of the other implementation will be simply ignored (as if no selection whatsoever).

mbuzdalov:
Ah, okay, I see where it goes: it shall be used something like:

```
new MultiComparator<DoubleSolution>(Arrays.asList(ranking.getSolutionComparator(),
        densityEstimator.getSolutionComparator()))
```

where we create `ranking` once and for the entire algorithm. In this case, it should work.

So what shall I do for jmetal5? Maybe just set the `DominanceRanking` property instead of a "custom" one, so that the rest of the world will understand the outcome of my code?

ajnebro:
In the NSGA-II implementation in the jmetal-auto package I don't use the `RankingComparator`. In fact, I have just added an implementation of SPEA2 (well, it is not exactly a version following the SPEA2 algorithm described in the literature) which is based on the same base class of NSGA-II (the  EvolutionaryAlgorithm class included in jmetal-auto.algorithm). The only differences relies in the ranking and density estimators: FastNonDominanceSortRanking and CrowdingDistanceDensityEstimator (NSGA-II), StrengthRanking and KnnDensityEstimator (SPEA2). 

In fact, both rankings and density estimators can be mixed in any combination.

ajnebro:
I don't know when I will have a jMetal 6.0 version ready, but I'm thinking that perhaps it's time to merge the jmetal6 branch into the master one.

The point is that my plan is to review everything, and most of the new ideas related to algorithm implementations I'm testing are now in the jmetal-auto branch. The drawback now is that there are different implementations of NSGA-II and now SPEA2, but for now I'm going to keep both of them.

mbuzdalov:
I did a rebase onto the new branch.

For this, I needed to update the `PointSolution` so that it can also contain constraints under the new interface.

I have also updated the method names following the changes in the `Ranking` interface and switched to the new way of `ConstraintHandling`.

I have also set my ranking class to use the same property identifier as `DominanceRanking`, so that they can indeed be used interchangeably in the older (non-auto) codebase.

mbuzdalov:
I have factored out the common code (-core and -auto) for both copies of this code.

They now shall use the same constraint handling mechanism, and the `Ranking` implementations are now different only in how they put ranks to the properties. So they shall behave identically, and there is now only one place to introduce changes (which is great).

mbuzdalov:
The code has been added and moved around by Antonio while refactoring, so now this PR is obsolete.

PR #382, opened 2019-12-10
Refine type parameters of a few classes, which removes certain warnings.

mbuzdalov:
`Ranking`, `DensityEstimator`, `Preference` receive "`extends Solution<?>`".

This removes a few type inference warnings, which were surprising at the first sight.

In `Preference` I have also fixed string comparison using `!=`, which is probably wrong.

### DIFF
--- a/jmetal-algorithm/src/test/java/org/uma/jmetal/algorithm/multiobjective/fame/FAMEIT.java
+++ b/jmetal-algorithm/src/test/java/org/uma/jmetal/algorithm/multiobjective/fame/FAMEIT.java
@@ -31,12 +31,12 @@ public class FAMEIT {
     int archiveSize=100 ;
     int maxEvaluations=25000;
 
-    algorithm = new FAME(problem,
+    algorithm = new FAME<>(problem,
             populationSize,
             archiveSize,
             maxEvaluations,
             selection,
-            new SequentialSolutionListEvaluator()) ;
+            new SequentialSolutionListEvaluator<>()) ;
 
     algorithm.run();
 
@@ -62,13 +62,13 @@ public class FAMEIT {
     int maxEvaluations = 25000;
 
     algorithm =
-        new FAME(
+        new FAME<>(
             problem,
             populationSize,
             archiveSize,
             maxEvaluations,
             selection,
-            new SequentialSolutionListEvaluator());
+            new SequentialSolutionListEvaluator<>());
 
     algorithm.run();
 
--- a/jmetal-auto/src/main/java/org/uma/jmetal/auto/util/preference/Preference.java
+++ b/jmetal-auto/src/main/java/org/uma/jmetal/auto/util/preference/Preference.java
@@ -2,6 +2,7 @@ package org.uma.jmetal.auto.util.preference;
 
 import org.uma.jmetal.component.densityestimator.DensityEstimator;
 import org.uma.jmetal.component.ranking.Ranking;
+import org.uma.jmetal.solution.Solution;
 
 import java.util.List;
 
@@ -11,7 +12,7 @@ import java.util.List;
  *
  * @author Antonio J. Nebro <antonio@lcc.uma.es>
  */
-public class Preference<S>  {
+public class Preference<S extends Solution<?>>  {
   private Ranking<S> ranking ;
   private DensityEstimator<S> densityEstimator ;
   private Preference<S> relatedPreference ;
@@ -48,7 +49,7 @@ public class Preference<S>  {
   }
 
   private boolean densityEstimatorsAreDifferent() {
-    return densityEstimator.getAttributeId() != relatedPreference.getDensityEstimator().getAttributeId() ;
+    return !densityEstimator.getAttributeId().equals(relatedPreference.getDensityEstimator().getAttributeId());
   }
 
   private void recomputeDensityEstimator() {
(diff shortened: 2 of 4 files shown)


matthieu-vergne:
> Ranking, DensityEstimator, Preference receive "extends Solution<?>".

I am not convinced this is a good move, since they obviously didn't need it. Otherwise they would already fail to compile.

There is an old discussion running on the progressive generalization of jMetal. It goes by pushing away the restrictions, possibly towards algorithms that do not even need a specific `Solution` concept in the first place, thus the alone `S`. Ranking for exemple is one of the class where I removed this useless extension. See this commit of May 10, 2017:
https://github.com/jMetal/jMetal/commit/daac65b0e07d4e293cc2efbbebfe72f4107b6ae0#diff-6d204b9dcdb07fda3d5ba1a598b69313

Consequently, this change is actually a regression. It has already been merged, but be aware that other people that may have used these classes and implemented them on their own (non solution) classes may have problems now. It is not a big deal, since it is easy to fix, but please don't increase the restrictions on a generics if the class defining it does not need it. This is the usage which should add restrictions as needed, not the definition.

> This removes a few type inference warnings, which were surprising at the first sight.

I don't think so. The addition of diamonds where you have generic classes missing them does resolve raw types warnings, but adding `extends Solution<?>` should either fix compilation problem (which were obviously absent) or add them (if it is too restrictive). Rather than adding these extensions, warnings are resolved by filling the raw types with proper types, either automatically with `<>` or manually with `<X>`.

> In Preference I have also fixed string comparison using !=, which is probably wrong.

Usually it is not, since the only way to have a "different" string despite the same value is to explicitly create a `String` object. More precisely:
- `"test" == "test"`
- `new String("test") != new String("test")`

Since the second is more or less never used, using `equals()` does not change much on that regard.

However, now we are no more `null`-friendly : if one of the value is `null`, instead of resulting in a boolean (`false` if the other is different, `true` if both are `null`) it will throw a NPE. Which is OK with me, since now we can use `Optional` to explicit when something is optional.

mbuzdalov:
> I am not convinced this is a good move, since they obviously didn't need it. Otherwise they would already fail to compile.

They tend to produce "the usage of a raw class" warnings at the use side.

>> This removes a few type inference warnings, which were surprising at the first sight.
> I don't think so. 

The NSGAII example shows this clearly:

```
    this.selectionOperator =
        new BinaryTournamentSelection<>(
            new MultiComparator<>(
                Arrays.asList(
                    ranking.getSolutionComparator(), densityEstimator.getSolutionComparator())));
```

Here, `ranking` is `Ranking<S>` which used to `extend Attribute` (a raw class), `densityEstimator` also traces back to the raw `Attribute`. The type of their `getSolutionComparator()` evaluates to a raw `java.util.Comparator`, which is clearly not the right thing to do. As a result, the type information is lost in the chain.

This pattern is not unique to NSGAII, of course.

> Since the second is more or less never used, using equals() does not change much on that regard.

When you run your code on a cluster, you find out that string interning does not work the way you expect any more. Within a single JVM, and with constant strings, "test" == "test". From two different JVMs, it is no longer true. I don't think you want so much surprises in your code.

> However, now we are no more null-friendly

If this is a concern, `Objects.equals` is our friend.

matthieu-vergne:
I didn't follow the many changes that occured lately, so I was not aware of the many moves. Since the old ones are still there and deprecated, the dependency argument surely falls.

After more investigations, I understand that you resolved the issues by fixing the raw type of `Attribute`, which itself required to introduce the extensions to fit the `Attribute` definition. Fair enough, since it resolves the warning. But it is not the extensions that solve the issue, this is the `Attribute` typing which does it. If you remove the extension from `Attribute` instead of adding it to its children, you resolve it too. The point is that `Attribute` has been designed as if requiring `Solution` but it doesn't. My move would have been to remove it from Attribute, not to add it to *all* its children.

The "keeping track" argument doesn't sound to me for a simple reason: it provides no useful information we already have. More precisely:
- we speak about solution everywhere, like in the method names, so we still know we deal with solutions there.
- we use the same generics `S` every time we deal with solutions, so we already keep track with that

This extensions provides not much information when you consider `S` and its surrounding. However, it does restrain the applicability of the classes we build, and it does so without any reason to do so. Moreover, generalization is a process we are going through since several years already. Restricting these classes means that other classes depending on them won't be further generalizable anymore. This generalization process goes iteratively, when we have time, and I think that introducing artificial restrictions is just adding more burden on ourselves, because the day we will need it we will have to redo it again.

There is a lot of work on that aspect in jMetal. So I would appreciate if we don't go in contradicting ways.

matthieu-vergne:
Raw types must be fixed, of course. I have fied hundreds of them already here and there are always others coming again. But useless extensions must be removed, not added.

mbuzdalov:
Did you mean this: #384?

matthieu-vergne:
Yes, exactly. Thank you for your effort. {^_°}b

The rationale is not really that "we don't want to constrain the user to extend their solutions from `Solution`", since for now someone cannot do much with jMetal without doing so. But it is for a long term generalization objective, to remove as much constraints as relevant to enlarge the applicable cases of jMetal. A lot of work remains before to start using other kinds of solutions, but this is undergoing work.

If you happen to find other cases like that, don't hesitate to relax their constraints too.

PR #384, opened 2019-12-10
Revert the type parameter narrowing.

mbuzdalov:
The concerns were raised that e.g. `Preference<S extends Solution<?>>` is too strict since we don't want to constrain the user to extend their solutions from `Solution`.

To follow that, but also to keep the type-safe path, we now drop the constraint on the `Attribute`, as this is entirely possible. Now everything typechecks again (e.g. no raw class usage in the related code), but this time it is less restrictive.

### DIFF
--- a/jmetal-auto/src/main/java/org/uma/jmetal/auto/util/preference/Preference.java
+++ b/jmetal-auto/src/main/java/org/uma/jmetal/auto/util/preference/Preference.java
@@ -2,7 +2,6 @@ package org.uma.jmetal.auto.util.preference;
 
 import org.uma.jmetal.component.densityestimator.DensityEstimator;
 import org.uma.jmetal.component.ranking.Ranking;
-import org.uma.jmetal.solution.Solution;
 
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


mbuzdalov:
@matthieu-vergne Could you please look at this?

ajnebro:
If I get the point, the idea is to allow the `Attribute` interface to be applied to other classes that do not to implement the `Solution` interface, isn't it? 

If so, I agree with the changes, although perhaps then the `org.uma.jmetal.solution.util` package could not be the most adequate one to include the `Attribute` interface.

mbuzdalov:
> If I get the point, the idea is to allow the Attribute interface to be applied to other classes that do not to implement the Solution interface, isn't it?

Basically yes. For its descendants, the similar constraint on the parameter had been already dropped previously, but this resulted in the inheritance from a raw `Attribute`, which is evil as it erased the type of a comparator.

Probably, the decision on where the `Attribute` shall reside may be postponed until jMetal has more descendants of `Attribute`.

PR #386, opened 2019-12-12
Speed-up of the MNDS

mbuzdalov:
In this branch, I speed up this repository's implementation of Merge Non-Dominated Sorting by some constant factor by writing the tightest loop better.

I have also added a simple benchmarking code which compares the performance of the mentioned implementation with its [equivalent in mbuzdalov/non-dominated-sorting](https://github.com/mbuzdalov/non-dominated-sorting/blob/master/implementations/src/main/java/ru/ifmo/nds/mnds/BitSetImplementation.java). The comparison results (on a certain fixed laptop) are as follows:

|  n  |  d  |  Moreno old  |  Moreno new  |  Buzdalov  |
|-----:|-----:|----------------|---------------|---------------|
| 10000 |  5  |    0.021     |    **0.019**     |    0.023   |
| 30000 |  5  |    0.148     |    **0.131**     |    0.186   |
| 100000 |  5  |    1.46      |    **1.20**     |    2.3     |
| 10000 | 10  |    0.023     |    0.023     |    **0.016**   |
| 30000 | 10  |    0.130     |    0.130     |    **0.114**  |
| 100000 | 10  |    1.02      |    1.01      |    **0.88**    |

Here, "Moreno old" is the state in master, "Moreno new" is the state of this PR, "Buzdalov" is the `BitSetImplementation` from my repo. The fastest result in each row is bold.

We can see that "Moreno new" strictly dominates "Moreno old". The comparison with "Buzdalov" is more involved: for dimension 5, the latter is worse, while for dimension 10, the latter is better. The exact reason is not yet known to me, but, most probably, this is due to `wordRanking`: for smaller dimensions it helps, for larger ones it is a useless overhead.

P.S.1: Here, "Buzdalov" is not the same as `ExperimentalFastNonDominanceRanking`. This is me having implemented a (so far simplified) MNDS on my own using stock bitsets.

### DIFF
--- a/jmetal-core/src/main/java/org/uma/jmetal/component/ranking/impl/util/MNDSBitsetManager.java
+++ b/jmetal-core/src/main/java/org/uma/jmetal/component/ranking/impl/util/MNDSBitsetManager.java
@@ -11,7 +11,6 @@ public class MNDSBitsetManager {
 	private final static int FIRST_WORD_RANGE = 0;
 	private final static int LAST_WORD_RANGE = 1;
 	private final static int N_BIT_ADDR = 6;
-	private final static int WORD_SIZE = 1 << N_BIT_ADDR;
 	private static final long WORD_MASK = 0xffffffffffffffffL;
 	private long[][] bitsets; //N bitsets. Each solution has a bitset
 	private int[][] bsRanges; //N ranges [first sol - last sol]. Each of each bitset of each solution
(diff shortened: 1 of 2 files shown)


matthieu-vergne:
If you are the author of something we publish, you should be part of the authors too (unless you refuse). So you will have your word on that. But if you are the one providing the code, just name the class as you see fit.

mbuzdalov:
> If you are the author of something we publish

I am not, at least as of now.

The P.S. on naming is just a piece of advice. However, since I like the algorithm, I feel I should do something, so that when this algorithm is attempted to be published, it will have as few obstacles as possible. And I surely think that the current name, "Merge Non-Dominated Sorting", as [on arxiv](https://arxiv.org/abs/1809.06106), is such an obstacle.

I would elaborate on that. The current name is based on the fact that the Moreno's implementation uses the mergesort algorithm to sort individuals on each objective. This does not make a good name, because:
* it does not rely on any properties of the mergesort algorithm, and any other stable sorting (and even a non-stable sorting!) can be used instead;
* it is not the bottleneck of the algorithm, so it does not contribute much to its performance.

So, to avoid the look-and-feel of a yet another to-be-rejected paper by "predatory researchers", I advice to come up with a more descriptive name, and have suggested one above.

mbuzdalov:
I have dropped all the comments on the name, since Antonio mentioned that the algorithm is nearly published in a journal with the present name.

PR #390, opened 2020-03-04
Fix of non-working budding in Coral Reefs Optimization

VEASTARK:
Budding (asexual reproduction or step 4 in the published paper)  does not work in the current implementation as no elements were added to the candidate list. I did a simple fix in this PR.

Reference:
Salcedo-Sanz, Sancho & Del Ser, Javier & Landa-Torres, Itziar & Gil-Lopez, Sergio & Portilla-Figueras, Antonio. (2014). The Coral Reefs Optimization Algorithm: A Novel Metaheuristic for Efficiently Solving Optimization Problems. TheScientificWorldJournal. 2014. 739768. 10.1155/2014/739768.


matthieu-vergne:
Good point. The size is the number of elements, which obviously starts at 0. The for loop should have built on the capacity instead. The sublist is a nice way to simplify it.

PR #394, opened 2020-03-19
Fix for PSO algorithms not updating their leading particle with a copy of the solution

VEASTARK:
Both StandardPSO algorithms do not use the copy function when updating their best found particle, leading to wrong results as the base particle will change in further iterations.


ajnebro:
Thanks a lot for the pull requests.

PR #406, opened 2020-07-03
Generalization

matthieu-vergne:
The goal is to generalize solution and operator implementations as much as possible, of course without breaking anything. Some stuff may be added to help. It will remain a draft as long as the content is not advanced enough to confirm a practical design.

Current target criteria, may evolve:
- Generic `Solution<T>` implemented
  - `BeanSolution<T>`
    - [x] Implemented
    - [x] Documented
    - [ ] Tested (WIP)
- Generic helpers implemented
  - `Metadata` interface + impl (replace `Solution` extensions like `DoubleSolution`)
    - [x] Implemented
    - [ ] Documented
    - [ ] Tested
- At least 3 different kinds of operators generalized
  - Operators on `DoubleSolution`
    - `DifferentialCrossoverVariation`
      - [x] `DoubleSolution` -> `Solution<Double>`
      - [ ] `Solution<Double>` -> `Solution<? extends Number>` (study)
    - `DifferentialEvolutionCrossover`
      - [x] `DoubleSolution` -> `Solution<Double>`
      - [ ] `Solution<Double>` -> `Solution<? extends Number>` (study)
    - `PolynomialMutation`
      - [x] `DoubleSolution` -> `Solution<Double>`
      - [ ] `Solution<Double>` -> `Solution<? extends Number>` (study)
  - Operators on something else
    - TBD
  - Others
    - `Evaluation`
      - [ ] Generalized
    - `Termination`
      - [ ] Generalized
- At least 3 different algorithms generalized
  - `FAME` = `DifferentialEvolutionCrossover` + `PolynomialMutation`
    - [ ] Generalized
  - `MOEADDE` = `DifferentialCrossoverVariation` + `DifferentialEvolutionCrossover` + `PolynomialMutation` + `Evaluation` + `Termination`
    - [x] Generalized to `Solution<Double>`
    - [ ] Generalized to `Solution<? extends Number>` (study)
  - `NSGAII` = `Evaluation` + `Termination`
    - [ ] Generalized
- At least 3 different problems generalized
  - TBD
  - TBD
  - TBD

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/abyss/ABYSSBuilder.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/abyss/ABYSSBuilder.java
@@ -45,7 +45,7 @@ public class ABYSSBuilder implements AlgorithmBuilder<ABYSS> {
     double distributionIndex = 20.0;
     this.crossoverOperator = new SBXCrossover(crossoverProbability, distributionIndex);
     double mutationProbability = 1.0 / problem.getNumberOfVariables();
-    this.mutationOperator = new PolynomialMutation(mutationProbability, distributionIndex);
+    this.mutationOperator = new PolynomialMutation<>(mutationProbability, distributionIndex);
     int improvementRounds = 1;
     this.archive = (CrowdingDistanceArchive<DoubleSolution>) archive;
 
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/cdg/CDG.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/cdg/CDG.java
@@ -35,7 +35,7 @@ import java.util.List;
  */
 @SuppressWarnings("serial")
 public class CDG extends AbstractCDG<DoubleSolution> {
-  private DifferentialEvolutionCrossover differentialEvolutionCrossover ;
+  private DifferentialEvolutionCrossover<DoubleSolution> differentialEvolutionCrossover ;
 
   public CDG(Problem<DoubleSolution> problem,
       int populationSize,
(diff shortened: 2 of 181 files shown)


ajnebro:
The idea of generalizing some components based on numerical types (such as polynomial mutation and SBX crossover) is nice. The [repairsolution](https://github.com/jMetal/jMetal/tree/master/jmetal-core/src/main/java/org/uma/jmetal/solution/util/repairsolution) package would also be a good candidate one.

However, in the case of Differential Evolution, this is intented to work with continuous variables (float or double), but not with integers, so I would keep it as is.

matthieu-vergne:
Thanks for the comment. The DE scope is one of the questions I intended to tackle in the study, since it requires calculation methods (+, -, *). I am glad you answer it already. But even if we remain on a continuous spectrum, I was thinking about BigDecimal, which would require to generalize to Number anyway. So it would be interesting to know if the DE is especially designed for values in [0;1], in which case it makes sense to stop at `Solution<Double>` only.

If you happen to have any other comments like that which can restrict the refactoring scope of the listed operators, please share. It will save me some valuable time to focus on the relevant parts.

I go through the operators required by the algorithms I refactor. And solution repair is part of MOEADDE, so it is indeed one of the expected candidates.

ajnebro:
> If you happen to have any other comments like that which can restrict the refactoring scope of the listed operators, please share. It will save me some valuable time to focus on the relevant parts.

Of course, it's not worth spending time on tasks that aren't necessary. I would focus on the aforementioned mutation, crossover and repair related classes. I wouldn't touch MOEA/D-DE. 

The repair has to do with the fact that, when varying variables whose values are enclosed into a range (e.g., [0.0, 1.0], [5, 15]),  the resulting value can be out of range. In those coses, the general strategy is to choose a random value in the desired range. What I did was to consider this as a particular case of a repairing strategy, which could adopt another approaches (for example, if the range is [1.3, 5.5] and a mutated value is 6.1, then it can take the upper limit 5.5, or alternative we could assign 1.3).

matthieu-vergne:
MOEADDE has already been generalized anyway. And it was worth it because it uses several of the most used operators, so it was a good case study. And since the final goal is to remove all the XxxSolution extensions, everything using them should be refactored anyway. The point is more to know *until* where should I refactor.

Thanks for the repairing explanation. I will check that.

If at some point you find some commits to be interesting to merge in master, please tell me that before. I will refactor the branch to make them available (git rebase & stuff).

PR #407, opened 2020-07-09
[Refactor] Replace constructors by factories

matthieu-vergne:
# Context

The various classes implemented in jMetal can be more or less complex.
This complexity reflects especially in the number of arguments needed to instantiate them.
To help the user in dealing with this complexity, jMetal classes often implement several constructors.
Each constructor may assign default values to unknown arguments or compute them from the others.

# Problems

Several problems can be highlighted regarding this practice.

Regarding the presentation, we can highlight several limitations:
- Constructors lack expressiveness: all constructors have the same name despite having different arguments.
It may not look like a problem when each constructor just reduces the number of arguments by assigning more default values.
However, when the arguments are simply different, we lack the reason why some arguments are relevant here and not there.
- Constructors must have different types of parameters.
Creating different instances which require the same types of arguments imposes to trick in some way the constructors.
For instance, by changing the order of the arguments to make it look like different types.
Of course, it is at the risk of tricking the user too.

Regarding the content, we can also highlight several limitations:
- Constructors do not allow to have code before another constructor call (`this(...)` or `super(...)`).
Any preliminary computation must be done directly in the call or moved to static methods.
In any case, if several arguments use the same computation, we cannot factor them without using heavy tricks.
- All constructors are strictly bound to the generics of their class.
If we want to generalize a class to expand the possible generic values, we must do so for all the constructors.
For example, generalizing an implementation from `Solution<Double>` to `Solution<? extends Number>`.
As such, constructors which assign default values must generalize these default values too.
In this example, operators on `Double` must be generalized to operators on `Number`.
Although it may be easy for some basic values, generalization can become infeasible or tricky.

# Solution

Rather than using constructors, we should use static factory methods.
Constructors should only require all the arguments, with no default value.
Ideally, there should have only one constructor.

For example, instead of having:
```java
new DifferentialEvolutionCrossover();
new DifferentialEvolutionCrossover(CR, F, DifferentialEvolutionCrossover.DE_VARIANT.RAND_1_BIN);
```

we have:
```java
DifferentialEvolutionCrossover.createDefault();
DifferentialEvolutionCrossover.createFromVariant(CR, F, DifferentialEvolutionCrossover.DE_VARIANT.RAND_1_BIN);
```

In this PR, I create the static factory methods but only deprecate the constructors.
This way nothing breaks and users get warnings to suggest updates.
However, I already do all the replacements in jMetal, so it is only useful for jMetal users.
If you think these constructors are not needed anymore, feel free to remove them or ask me to do so here.

I apply this pattern to the following classes:
- DifferentialCrossoverVariation
- DifferentialEvolutionCrossover
- PolynomialMutation
- MOEADDE

The PR is split in different commits for readability, so each class as a commit to:
- create its static factory methods
- replace its constructor calls

If you review the changes, you can do it by commit to keep it simple.

The focus on these classes is because I am working on them for the generalization process (#406).
By accepting these changes, I will be able to consider the constructors as relic of the past.
This way, I can just use simple tricks on them rather than add complexity to generalize them properly.
If we remove them, I won't have to care about them at all.

Other classes are expected to be generalized in #406.
If this PR is accepted, expect future PRs for other classes as well.

# Discussion

Static factory methods do solve the problems highlighted:
- Static factory methods can be named as needed, as opposed to constructors.
- Because of that, we can have the same types of parameters as much as we want.
We only need to choose a different name, preferably which highlight the actual difference.
- Static factory methods can do all the computation they want before to instantiate the class.
- Because of being static, they are not bound to the generic value of the class.
Each static factory method can specialize on a specific generic value.
When generalizing the class, the existing static factory methods can just remain how they are and new ones be added.

Additionally, if there are many static factory methods, you can create a separate factory class to lighten the class.
If the factory methods combinations are numerous, it can naturally evolve to a builder class.

However, there is a limit: the instances returned by static factory methods cannot be overriden.
When calling a constructor, you can also decide to override some of its methods at the same time.
But once the static factory method returns, the instance is already created.
But the same result can be achieved by using wrapper classes, like [decorators](https://en.wikipedia.org/wiki/Decorator_pattern) or [proxies](https://en.wikipedia.org/wiki/Proxy_pattern).
So it is not a blocking limitation, rather a different habit.

Although different, static factory methods can be compared to the [abstract factory pattern](https://en.wikipedia.org/wiki/Abstract_factory_pattern).
We could say it is like a "non abstract" abstract factory pattern, since we implement them immediately.
However, it should not be mixed with the [factory method pattern](https://en.wikipedia.org/wiki/Factory_method_pattern).
This one is about providing *to the class* a factory method to *instantiate a component* used in the class.
Here we provide a way to *instantiate the class* to *the user*, so we go the other way around.

You can find a more extensive description of the static factory methods in [an article of Joshua Bloch](https://www.drdobbs.com/jvm/creating-and-destroying-java-objects-par/208403883).
Some points are not relevant anymore (it was written for Java 6), but it draws additional advantages and inconvenients.

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/abyss/ABYSSBuilder.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/abyss/ABYSSBuilder.java
@@ -45,7 +45,7 @@ public class ABYSSBuilder implements AlgorithmBuilder<ABYSS> {
     double distributionIndex = 20.0;
     this.crossoverOperator = new SBXCrossover(crossoverProbability, distributionIndex);
     double mutationProbability = 1.0 / problem.getNumberOfVariables();
-    this.mutationOperator = new PolynomialMutation(mutationProbability, distributionIndex);
+    this.mutationOperator = PolynomialMutation.createWithDoubleDefaults(mutationProbability, distributionIndex);
     int improvementRounds = 1;
     this.archive = (CrowdingDistanceArchive<DoubleSolution>) archive;
 
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/cdg/CDGBuilder.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/cdg/CDGBuilder.java
@@ -61,7 +61,7 @@ public class CDGBuilder implements AlgorithmBuilder<AbstractCDG<DoubleSolution>>
     populationSize = 300;
     resultPopulationSize = 300;
     maxEvaluations = 300000;
-    crossover = new DifferentialEvolutionCrossover();
+    crossover = DifferentialEvolutionCrossover.createDefault();
     neighborhoodSelectionProbability = 0.9;
     numberOfThreads = 1;
     sigma_ = 10e-6;
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/fame/FAME.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/fame/FAME.java
@@ -185,7 +185,7 @@ public class FAME<S extends DoubleSolution> extends SteadyStateNSGAII<S> {
     double probabilityPolynomial, DristributionIndex;
     probabilityPolynomial = 0.30;
     DristributionIndex = 20;
-    PolynomialMutation mutationPolynomial = new PolynomialMutation(probabilityPolynomial, DristributionIndex);
+    PolynomialMutation mutationPolynomial = PolynomialMutation.createWithDoubleDefaults(probabilityPolynomial, DristributionIndex);
 
     double probabilityUniform, perturbation;
     probabilityUniform = 0.30;
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/fame/FAME.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/fame/FAME.java
@@ -196,7 +196,7 @@ public class FAME<S extends DoubleSolution> extends SteadyStateNSGAII<S> {
     CR = 1.0;
     F = 0.5;
     DifferentialEvolutionCrossover crossoverOperator_DE =
-        new DifferentialEvolutionCrossover(CR, F, DifferentialEvolutionCrossover.DE_VARIANT.RAND_1_BIN);
+        DifferentialEvolutionCrossover.createFromVariant(CR, F, DifferentialEvolutionCrossover.DE_VARIANT.RAND_1_BIN);
 
     double crossoverProbability, crossoverDistributionIndex;
     crossoverProbability = 1.0;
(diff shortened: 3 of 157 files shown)


matthieu-vergne:
The Codecov report indicates a significant increase in misses (now is +83). This is because the factory methods duplicate the constructors but the tests have only been moved to the factory methods, so the constructors are not covered anymore. Since they are deprecated, this increase should be recovered when the deprecated code will be removed in future releases.

ajnebro:
Matthieu, this is a really really interesting issue. Let me think about it and check what you have implemented. 

My idea with the jMetal 6 release is to re-think everything, so this could be a point. In a private branch, I'm trying different things: switch to jUnit 5, remove the jMetal core classes dependency of some components (e.g., the quality indicators can take as a parameter a matrix of doubles instead of a List<Solution>, what could facilitate to use them in applications not relying on the jMetal code), etc. What I don't know is when the release will be ready, but I'm not in a hurry ;).

matthieu-vergne:
> My idea with the jMetal 6 release is to re-think everything

Why not, although I would go in a progressive way. It is usually possible although it requires some discipline and I am willing to help on that if required.

> switch to jUnit 5

I thought about doing that too since now I only produce jUnit 5 tests.
There is some tricks with mocks, like if you use specific runners it may need some refactoring first.
I can definitely help on that.

> remove the jMetal core classes dependency of some components (e.g., the quality indicators can take as a parameter a matrix of doubles instead of a List, what could facilitate to use them in applications not relying on the jMetal code), etc.

Seems to fit well in a generalization process. {^_^}b

> What I don't know is when the release will be ready, but I'm not in a hurry ;).

As I said, I prefer to make it progressive, deprecating the old stuff and all.
Not only for the users, but also for us, because breaking stuff imposes to fix all the broken stuff at once before to make any release.
And it facilitates the parallel contributions.
Otherwise one is working on a version while the other is working on another, and it makes it hard to integrate.

As a side note, I have named the factory methods based on what I could grasp of their specifics. Feel free to think about new names.

matthieu-vergne:
@ajnebro What happened? Did you remove all the files?

ajnebro:
I reorganized some packages. The `jmetal-lab/src/main/java/org/uma/jmetal/lab` package is now part of `jmetal-lab/src/main/java/org/uma/jmetal/experiment/`, and the algorithms in `jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective` have been moved to jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/jmetal5version`,

matthieu-vergne:
> the algorithms in `.../singleobjective` have been moved to `.../singleobjective/jmetal5version`

Keeping old and new code together is usually a (really) bad move.
Because if the old code depends on some maintained code, we will have to take care of both old and new constraints, which usually brings more costs than advantages.
If the point is to get rid of them, there is no point for me to work on them.

The usual way to go is to keep a legacy branch with the old code while going ahead with the new code on master and getting rid of the old code there.
If something needs to be done with the old code, it is done on the legacy branch only.
And you don't have to cope with both legacy and new constraints at the same time.

Moving them to a new folder breaks everything anyway.
So if you have to break everything, don't keep constraints that can be avoided.

Really, seeing these jmetal5 folders scares me as a maintainer.
I would prefer you to tell me that I should throw everything and restart from scratch on the new code rather than having to think that now I will have to cope with both legacy and new.

ajnebro:
The issue is that I'm exploring new ideas (such as a new template for evolutionary algorithtms), but in no case I'm able to start another jMetal project from scratch. So, my approach was to keep all the jMetal 5 code, and add new versions of some algorithms in what it would the next 6.0 release. That's the reason of having some `jmetal5version` packages.

Atfer reading your comments I see now that that approach is wrong, because it leads to confusion. During this week I'm going to restore the jMetal 5 branch in https://github/jMetal/jMetal and start a new project in https://github/jMetal/jMetal6, which willl use the former as a dependency.

matthieu-vergne:
Creating a new project may be overkill, but way safer indeed.
Then I will switch to the new project. {^_°}b

Issue #411, opened 2020-07-23
Solution refactor


matthieu-vergne:
After discussing it in other threads, here is a plan to improve `Solution` in an iterative way. Basically, it is a sequence of refactoring and deprecation tasks, without removing anything to not break existing code, including for the jMetal users.

Each deprecated element must be documented with the `@deprecated` Javadoc tag to tell how to migrate. Since we have several deprecation steps, it may be that an old deprecation suggest to migrate to another (more recent) deprecated stuff. That is not an issue because this newly deprecated stuff should have its own suggestion. The user should eventually reach a non-deprecated component after some migration steps.

Some steps of this plan, like creating a class, are atomic: they are done in one shot. Other steps, like replacing methods, are way heavier but can be done progressively, like few methods at a time. This is what makes the whole plan progressive: it can be stopped and resumed at any point.

# Generalyze from the many `XxxSolution` extensions to the unique `Solution<Xxx>`

The first step is to simplify the types hierarchy by going back to the generic interface `Solution`. We can rely on its attributes to store additional data:
- [ ] Create an interface like this one (although I would probably rename it `Property` or alike) to read/write additional values from/in solutions : https://github.com/jMetal/jMetal/commit/ed79bba42a9bf11a83aa347eb9a1072e870b3372
- [ ] Provide an implementation for each additional property brought by extensions of `Solution`, like I did here for `DoubleSolution`: https://github.com/jMetal/jMetal/commit/1215981aaaa534fd37aa7b4b16c21ed37ee6bbac
- [ ] Progressively refactor all the components to use these objects instead of the specific methods of the interfaces, using these implementations by default. As discussed earlier, I will favor factory methods over constructors.
- [ ] After having no specific dependency anymore, replace all the `XxxSolution` by the more generic `Solution<Xxx>`

# Allow the use of custom solution generators

At this point, we don't need the extensions anymore, but we are still generating solutions from the solutions themselves (copy) and from problems (createSolution), so these extensions are still in use. We must first allow their replacement:
- [ ] Generalize the no-arg instantiators (random, problem) by `Supplier` objects, with the current components (random, problem) used as default. Again, I will favor factory methods over constructors.
- [ ] Generalize the copies, which take another solution in argument, by `UnaryOperator<Solution<T>>`, maybe making our on interface for readability, like `SolutionCopier<T>.copy(solution)`. We can use `Solution.copy()` as default again.

# Use of a generic `Solution` implementation

[...]

PR #413, opened 2020-07-27
Generalyze Solution extensions

matthieu-vergne:
Aim to address the first section of #411 on a specific `Solution` extension. Here are the available extensions and the number of references found by Eclipse:
- BinarySolution (502 references)
- DoubleSolution (3 563 references)
- IntegerSolution (485 references)
- IntegerDoubleSolution [deprecated] (22 references)
- PermutationSolution (85 references)
- SequenceSolution (4 references)

I don't ignore IntegerDoubleSolution although it is already deprecated.
We will need to confirm that nothing active (not deprecated) uses it.

The references do not count the references of the specific implementations.
It looked to me that we mainly use interfaces, so let's rely on that here.

Since they all add 1 or 2 methods, they are comparable in terms of refactoring efforts.
The number of references thus gives a reliable idea of their priority.
Since DoubleSolution is by far the most used, I will start by this one to cover as much use cases as possible.

The first step is to simplify the types hierarchy by replacing `DoubleSolution` by the generic interface `Solution<Double>`. We can rely on its attributes to store additional data:
- [x] Create a Metadata interface to read/write additional values from/in solutions, like this one : https://github.com/jMetal/jMetal/commit/ed79bba42a9bf11a83aa347eb9a1072e870b3372
- [x] Provide a Metadata implementation for `DoubleSolution.getBounds()`, like here for legacy lower/upper bounds: https://github.com/jMetal/jMetal/commit/1215981aaaa534fd37aa7b4b16c21ed37ee6bbac
- [ ] Progressively refactor all the components to use the Metadata objects instead of the `Solution` method, using these implementation by default. As discussed earlier, I will favor factory methods over constructors.
- [ ] After having no specific dependency anymore, replace all the `DoubleSolution` by the more generic `Solution<Double>`

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/particleswarmoptimization/StandardPSO2007.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/singleobjective/particleswarmoptimization/StandardPSO2007.java
@@ -8,6 +8,7 @@ import org.uma.jmetal.solution.doublesolution.DoubleSolution;
 import org.uma.jmetal.util.bounds.Bounds ;
 import org.uma.jmetal.util.comparator.ObjectiveComparator;
 import org.uma.jmetal.util.evaluator.SolutionListEvaluator;
+import org.uma.jmetal.util.metadata.Metadata ;
 import org.uma.jmetal.util.neighborhood.impl.AdaptiveRandomNeighborhood;
 import org.uma.jmetal.util.pseudorandom.JMetalRandom;
 import org.uma.jmetal.util.solutionattribute.impl.GenericSolutionAttribute;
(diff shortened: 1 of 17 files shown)


ajnebro:
> I don't ignore IntegerDoubleSolution although it is already deprecated.

You can ignore that class, as it is replaced by `CompositeSolution`, which is a collection of solutions of any type. 

Any alternative to the name `CompositeSolution` would be welcome ;)

matthieu-vergne:
Actually seems to me the best name, since it makes think about the composite design pattern:
https://en.wikipedia.org/wiki/Composite_pattern

As long as it fits the pattern, I would recommend to use this name.

For IntegerDoubleSolution, I must confirm that it has been properly replaced everywhere. I don't plan to refactor it more than necessary (to make the code compile).

PR #414, opened 2020-07-27
Factor bounds

matthieu-vergne:
We basically always call them together:
```java
solution.getLowerBound(i)
solution.getUpperBound(i)
```
Depending on the implementation, it means we systematically do twice the search.
Since we always ask them together, let's just put them together and search once:
```java
bounds = solution.getBounds(i);
bounds.getLowerBound()
bounds.getUpperBound()
```

This PR does exactly that (and deprecates the two legacy methods) for the two solution interfaces concerned with this issue:
- `DoubleSolution`
- `IntegerSolution`

It also covers the related bounded problems.

Since lower and upper bounds are stored in `Pair` in some places, we replace them by proper `Bounds` to avoid conversion overhead. Again, constructors and getters based on `Pair` are deprecated and replaced by ones using `Bounds`.

All use the same generic interface which can be applied to any `Comparable` type:
```java
interface Bounds<T extends Comparable<T>>
```
so feel free to reuse if you have bounds elsewhere, like for `RepairDoubleSolution` which also use lower and upper bounds.

A coincidental advantage is that it simplifies the refactoring for #411 (more specifically #413 and later PRs).

### DIFF
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/abyss/ABYSS.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/abyss/ABYSS.java
@@ -9,6 +9,7 @@ import org.uma.jmetal.solution.doublesolution.DoubleSolution;
 import org.uma.jmetal.util.SolutionUtils;
 import org.uma.jmetal.util.archive.Archive;
 import org.uma.jmetal.util.archive.impl.CrowdingDistanceArchive;
+import org.uma.jmetal.util.bounds.Bounds ;
 import org.uma.jmetal.util.comparator.CrowdingDistanceComparator;
 import org.uma.jmetal.util.comparator.DominanceComparator;
 import org.uma.jmetal.util.comparator.EqualSolutionsComparator;
@@ -151,11 +152,11 @@ public class ABYSS extends AbstractScatterSearch<DoubleSolution, List<DoubleSolu
       frequency[range][i]++;
       sumOfFrequencyValues[i]++;
 
-      double low = ((DoubleProblem)problem).getLowerBound(i) + range *
-          (((DoubleProblem)problem).getUpperBound(i) -
-              ((DoubleProblem)problem).getLowerBound(i)) / numberOfSubRanges;
-      double high = low + (((DoubleProblem)problem).getUpperBound(i) -
-          ((DoubleProblem)problem).getLowerBound(i)) / numberOfSubRanges;
+      Bounds<Double> bounds = ((DoubleProblem)problem).getBoundsForVariables().get(i) ;
+      Double lowerBound = bounds.getLowerBound() ;
+      Double upperBound = bounds.getUpperBound() ;
+      double low = lowerBound + range * (upperBound - lowerBound) / numberOfSubRanges ;
+      double high = low + (upperBound - lowerBound) / numberOfSubRanges ;
 
       value = randomGenerator.nextDouble(low, high);
       solution.setVariable(i, value);
--- a/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/dmopso/DMOPSO.java
+++ b/jmetal-algorithm/src/main/java/org/uma/jmetal/algorithm/multiobjective/dmopso/DMOPSO.java
@@ -4,6 +4,7 @@ import org.uma.jmetal.algorithm.Algorithm;
 import org.uma.jmetal.problem.doubleproblem.DoubleProblem;
 import org.uma.jmetal.solution.doublesolution.DoubleSolution;
 import org.uma.jmetal.util.JMetalException;
+import org.uma.jmetal.util.bounds.Bounds ;
 import org.uma.jmetal.util.evaluator.SolutionListEvaluator;
 import org.uma.jmetal.util.evaluator.impl.SequentialSolutionListEvaluator;
 import org.uma.jmetal.util.pseudorandom.JMetalRandom;
@@ -122,8 +123,8 @@ public class DMOPSO implements Algorithm<List<DoubleSolution>> {
     deltaMax = new double[problem.getNumberOfVariables()];
     deltaMin = new double[problem.getNumberOfVariables()];
     for (int i = 0; i < problem.getNumberOfVariables(); i++) {
-      deltaMax[i] = (problem.getUpperBound(i) -
-              problem.getLowerBound(i)) / 2.0;
+      Bounds<Double> bounds = problem.getBoundsForVariables().get(i) ;
+      deltaMax[i] = (bounds.getUpperBound() - bounds.getLowerBound()) / 2.0 ;
       deltaMin[i] = -deltaMax[i];
     }
   }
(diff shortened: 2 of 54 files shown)


matthieu-vergne:
@ajnebro I postpone the solution refactoring to the closing of this PR, since it depends on whether or not it is accepted.

matthieu-vergne:
@ajnebro This refactoring is done. Please consider when you have some time. I have 1 commit for the main stuff, and 1 commit for its application everywhere.

ajnebro:
I have some concerns about the `Bounds` interface. In the example below, we are replacing two calls:

```java
solution.getLowerBound(i)
solution.getUpperBound(i)
```

by creating an object and another two calls: 

```java
bounds = solution.getBounds(i);
bounds.getLowerBound()
bounds.getUpperBound()
```

I don't see now any advantage in this case. 

About the `restrict()` method, I like the idea, but assigning the bounds when a value is out of them is only a possible strategy. In this sense, I included the [RepairDoubleSolution](https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/solution/util/repairsolution/RepairDoubleSolution.java) interface with three implementations: https://github.com/jMetal/jMetal/tree/master/jmetal-core/src/main/java/org/uma/jmetal/solution/util/repairsolution.  This approach is a first attempt, and any improvement around it would be nice.

We must keep in mind the context of the bounded variables. When using solutions having N integer or real bounded variables, performance is not an important issue when N is small (i.e, less than 500 or 1000), but a trend now is large scale optimization, where N can take values around tens of millions. In these cases, efficiency (in terms of storage and computing) matters a lot.

matthieu-vergne:
> I have some concerns about the `Bounds` interface. In the example below, we are replacing two calls:
> 
> ```java
> solution.getLowerBound(i)
> solution.getUpperBound(i)
> ```
> 
> by creating an object and another two calls:
> 
> ```java
> bounds = solution.getBounds(i);
> bounds.getLowerBound()
> bounds.getUpperBound()
> ```
> 
> I don't see now any advantage in this case.

1. You replace two *search* calls (not necessarily O(1) depending on implementation) by 1 *search* call + 2 *access* calls (always O(1)), so algorithmically you gain in performance. If the search is done through an array, it is equivalent to an access, so you replace 2 calls by 3, but there is hundreds of access calls in a solution processing, adding 1 doesn't change much.
2. Access calls can be optimised by the JVM through method inlining, so you can basically ignore all access calls, which means we reduce 2 search calls to 1.
3. The algorithm becomes more robust to dynamic changes: you retrieve the context in 1 call, so you avoid issues like getting the lower bound first, then having a change of the bounds in a concurrent processing, and then getting the upper bound, which can lead to inconsistent bounds. I don't know (and I don't care) whether it is relevant for this or that implementation, the point is that I don't have to ask this question anymore if I guarantee that I retrieve all the context in one call.
4. By keeping them tightly coupled in a single object, you allow yourself to provide common computations on that object directly, like `restrict()`. This would allow me to define a single Metadata for `Bounds` instead of 1 for each bound separately in #413.

> About the `restrict()` method, I like the idea, but assigning the bounds when a value is out of them is only a possible strategy. In this sense, I included the [RepairDoubleSolution](https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/solution/util/repairsolution/RepairDoubleSolution.java) interface with three implementations: https://github.com/jMetal/jMetal/tree/master/jmetal-core/src/main/java/org/uma/jmetal/solution/util/repairsolution. This approach is a first attempt, and any improvement around it would be nice.

Of course it is. The current `restrict(value)` is the most simple case, where in the event of an unsatisfied lower/upper bound, we have a predefined behaviour of using the corresponding bound. The method can be generalized to allow further behaviours, like :
- `restrict(value, replacement)` if we want to replace the value by a constant when it is out
- `restrict(value, lower replacement, upper replacement)` if we want to replace the value by a constant for each case
- `restrict(value, value -> replacement logic)` if we want to compute the replacement value based on the current one
- `restrict(value, value -> lower logic, value -> upper logic)` if we want to separate lower and upper cases behaviours

Considering your repair implementations:
- `RepairDoubleSolutionWithBoundValue` can be replaced by the current `restrict(value)`.
- `RepairDoubleSolutionWithOppositeBoundValue` can be replaced by `restrict(value, lower, upper)`
- `RepairDoubleSolutionWithRandomValue` can be replaced by `restrict(value, value -> replacement logic)`

Anyway, all are just a more specific case for `restrict(value, value -> lower logic, value -> upper logic)`, so we cover a rather extensive set of behaviours with this method.

Meanwhile, the checks (e.g. lower < upper) have been already made when creating the `Bounds` object, so you really don't need them for each repair action. Since the only thing remaining is the processing, contained in the corresponding `restrict` method, all you need when defining a repair action is to provide the right `restrict` method to use. No need for a specific repairing class. Although you may create dedicated methods in a factory class (to no overload `Bounds` with many factory methods) if you really want to.

> We must keep in mind the context of the bounded variables. When using solutions having N integer or real bounded variables, performance is not an important issue when N is small (i.e, less than 500 or 1000), but a trend now is large scale optimization, where N can take values around tens of millions. In these cases, efficiency (in terms of storage and computing) matters a lot.

In your current code, you already store 1 container of `Pair` instances. I just replace them by `Bounds` instances, so it is the same in terms of space. In terms of computation time, I already covered that point above: you reduce 2 searches (in the same container) to 1.

matthieu-vergne:
In short, you currently use `Pair` to store the bounds, but don't expose it everywhere by retrieving the bounds separately. This is good because we avoid having an external dependency everywhere in our code, but bad because you duplicate the search calls. I just gave this object an internal representation `Bounds` so we can expose it everywhere and factor the search calls.

matthieu-vergne:
And as an additional point, in Java 14 (experimental) and probably Java 17 (release), `Bounds` should be representable as a `Record`:
- tutorial : https://www.baeldung.com/java-record-keyword
- doc : https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/lang/Record.html
- JEP : https://openjdk.java.net/jeps/359

We may exploit further optimization already integrated in the JVM when using them.

ajnebro:
Let's go then :)

matthieu-vergne:
\\{^o^}/

ajnebro:
About the need of repairing strategies, just take a look at the [PolynomialMutation](https://github.com/jMetal/jMetal/blob/master/jmetal-core/src/main/java/org/uma/jmetal/operator/mutation/impl/PolynomialMutation.java) class.

The issue is that the strategy for repairing (or restricting) variables is a class parameter:

https://github.com/jMetal/jMetal/blob/069f4d05432e5f66a069c7fc8c5f29d127f25e73/jmetal-core/src/main/java/org/uma/jmetal/operator/mutation/impl/PolynomialMutation.java#L70-L78

so, in the implementation, we can use the indicated repairing strategy as shown here in line 165:

https://github.com/jMetal/jMetal/blob/069f4d05432e5f66a069c7fc8c5f29d127f25e73/jmetal-core/src/main/java/org/uma/jmetal/operator/mutation/impl/PolynomialMutation.java#L139-L172

matthieu-vergne:
RepairDoubleSolution is an interface with 1 method : you can use lambdas to feed such a parameter. For example this:
```java
  /** Constructor */
  public PolynomialMutation(double mutationProbability, double distributionIndex) {
    this(mutationProbability, distributionIndex, new RepairDoubleSolutionWithBoundValue());
  }
```
can be replaced by this:
```java
  /** Constructor */
  public PolynomialMutation(double mutationProbability, double distributionIndex) {
    this(mutationProbability, distributionIndex, (value, lowerBound, upperBound) -> Bounds.create(lowerBound, upperBound).restrict(value));
  }
```
So you can already technically replace it.

It is not hard then to think about deprecating the current interface and related constructors, which require separate numbers, to add new ones using Bounds to avoid conversion overheads.

PR #460, opened 2023-04-02
Use constructor arg to get number of variables in UF9

dhadka:
Calling the UF9 constructor results in a NullPointerException because it's calling the function `numberOfVariables()` which requires the `bounds` to be set.  Instead, use the argument `numberOfVariables`.

```
    [junit] Cannot invoke "java.util.List.size()" because "this.bounds" is null
    [junit] java.lang.NullPointerException: Cannot invoke "java.util.List.size()" because "this.bounds" is null
    [junit] 	at org.uma.jmetal.problem.doubleproblem.impl.AbstractDoubleProblem.numberOfVariables(AbstractDoubleProblem.java:27)
    [junit] 	at org.uma.jmetal.problem.multiobjective.UF.UF9.<init>(UF9.java:41)
    [junit] 	at org.uma.jmetal.problem.multiobjective.UF.UF9.<init>(UF9.java:20)
```


ajnebro:
Thanks for reporting the bug.

Antonio

